Issue #168, opened 2018-08-27
User Guide - GPUP add examples


mlucinhdl:
Add examples

mlucinhdl:
Fixes merged in #197

PR #197, opened 2018-08-29
Iss168

bsimeunovichdl:
GPUP - changed tests comments and printing one error in action.cpp

### DIFF
--- a/gpup.so/src/action.cpp
+++ b/gpup.so/src/action.cpp
@@ -246,6 +246,7 @@ void action::property_get_value(string gpu_id, int node_id) {
                         // log the error
                         msg = action_name + " " + MODULE_NAME + " " +
                         JSON_CREATE_NODE_ERROR;
+                        cerr << "RVS-GPUP: " << msg;
                     }
                 }
             }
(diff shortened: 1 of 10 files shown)


gpup.so/src/action.cpp
```diff
@@ -247,7 +246,6 @@ void action::property_get_value(string gpu_id, int node_id) {
                         // log the error
                         msg = action_name + " " + MODULE_NAME + " " +
                         JSON_CREATE_NODE_ERROR;
-                        cerr << "RVS-GPUP: " << msg;
```
mlucinhdl: This line is needed, please fix

gpup.so/src/action.cpp
```diff
@@ -382,7 +379,6 @@ int action::run(void) {
     rvs::actionbase::property_get_action_name(&error);
     if (error == 2) {
       msg = "action field is missing in gst module";
-      cerr << "RVS-GPUP: " << msg;
```
mlucinhdl: This line is needed

gpup.so/src/action.cpp
```diff
@@ -412,7 +408,6 @@ int action::run(void) {
             // log the error
             msg = action_name + " " + MODULE_NAME + " " +
             JSON_CREATE_NODE_ERROR;
-            cerr << "RVS-GPUP: " << msg;
```
mlucinhdl: This line is needed

mlucinhdl:
Fixes for #168

Issue #169, opened 2018-08-27
User Guide - GM add examples


mlucinhdl:
Add examples.

mlucinhdl:
Fixes merged in #216

PR #216, opened 2018-08-30
Fix for #169

mlucinhdl:


mlucinhdl:
Added examples for GM module.

Issue #287, opened 2018-10-18
PQT module ignores duration key


jedwards-AMD:
The PQT module is ignoring the duration key, and will only perform a transfer test for as long as it takes for the transfer to finish. This makes the log_interval key useless and doesn't allow the user to determine the sustained data-transfer rates between host and device. Consider this configuration file:
actions:
- name: action_1
  device: all
  duration: 100000
  module: pqt
  log_interval: 1000
  peers: all
  test_bandwidth: true
  bidirectional: true
  block_size: 1000000 
  .
 According to the RVS specification this should result in 100 INFO messages giving the average bandwidth over a 1 second window (the log_interval) for each device being measured. Instead, it only gives one RESULT line per device. The result is required output, but the time series of the test is never executed.

mlucinhdl:
Considered approved.

Issue #286, opened 2018-10-18
PEBB module ignores duration


jedwards-AMD:
The PEBB module is ignoring the duration key, and will only perform a transfer test for as long as it takes for the transfer to finish. This makes the log_interval key useless and doesn't allow the user to determine the sustained data-transfer rates between host and device. Consider this configuration file:
actions:
- name: h2d-d2h
  device: all
  duration: 100000
  module: pebb
  log_interval: 1000
  device_to_host: true
  host_to_device: false
  block_size: 10240000
.
According to the RVS specification this should result in 100 INFO messages giving the average bandwidth over a 1 second window (the log_interval) for each device being measured. Instead, it only gives one RESULT line per device. The result is required output, but the time series of the test is never executed.

mlucinhdl:
Please try running with -d 3 switch to allow INFO messages to be displayed.

jedwards-AMD:
I have used the -d 3 option. Moreover I have attempted to run using variable durations. Do you have a config script that will run with longer duration than a single copy? Maybe this is an issue with parameter passing.

mlucinhdl:
It seems you are missing `parallel: true`
When `parallel` is missing it is looking for `count` . If not found, the default is 1 and test exits quickly.

mlucinhdl:
Just a small clarification: if `parallel:false` test will run until `duration` expires of number of **test transfers** reaches `count` (whichever comes first).
We may change interpretation of `count` so that it repeats test (as if there were multiple actions in .conf file)

As per duration I have to double check.

mlucinhdl:
Considered approved.

PR #290, opened 2018-10-29
Iss286

mlucinhdl:
Fixes for #286, #287


jedwards-AMD:
The duration seems to be working better with the latest build, but the parallel flag still seems strange for this. If I specify 1 device using a 10 second interval for the run, I get a 10 second run and the report of a 10 second duration in the RESULT line. If I specify two devices and set the parallel flag to 'true' I get a test that runs for 10 seconds on both devices and a RESULT line that specifies about a 5 second duration per device. The same is true if parallel is set to 'false', and the message interleaving is the same.
.
What I would expect to see is that if parallel was 'true' I would get runs that were host to device occurring at the same time between the devices and the host CPUs, INFO messages would be interleaved and the duration for each test would be 10 seconds. The total duration would also be 10 seconds. If the parallel was false I would see the tests occur one after the othere, INFO messages wouldn't be interleaved and the test duration would be 10 seconds for each device, but the total duration would be 10 * num devices.

PR #430, opened 2019-10-19
Modify cmakelist to work with librocm_smi64.so from ROCm2.9

sunway513:
This PR is to update the librocm_smi64 version config to be compatible with ROCm2.9.


rakeshsv:
No

PR #434, opened 2019-10-31
Edited exact path for rocm_smi and ROCT

pramenku:
Below changes was done.
  set(ROCM_SMI_INC_DIR "/opt/rocm/rocm_smi/include")
  set(ROCM_SMI_LIB_DIR "/opt/rocm/rocm_smi/lib/")
  set(ROCT_INC_DIR "/opt/rocm/include"


CMakeLists.txt
```diff
@@ -105,15 +105,15 @@ set(RVS_BUILD_TESTS TRUE CACHE BOOL "TRUE if tests are to be built")
 
 set(RVS_DO_TRACE "1" CACHE STRING "Expand RVSTRACE_ macro")
 set(RVS_ROCBLAS "0" CACHE STRING "1 = use local rocBLAS")
-set(RVS_ROCMSMI "1" CACHE STRING "1 = use local rocm_smi_lib")
+set(RVS_ROCMSMI "0" CACHE STRING "1 = use local rocm_smi_lib")
```
rakeshsv: We want to use the local SMI isnt it? Making it zero , we cant use the local one isnt it? My thinking is use the local SMI until this tool gets merged into the docker container. 

If its zero its depending on the ROCM installation and other dependencies isnt it?  If they are not installed then we are in trouble I think

pramenku: It's good to be using rocm_smi64 packaged with rocm release like we are using rocblas and made zero for rocblas in file. To run RVS, anyway we need to install rocm.
ROCm driver always install rocm_smi64 by default. 
It will help to test our package also which we provide with rocm release

We are planning to package RVS as rocm package. Docker container method is not required.

rakeshsv:
Merge pull request #434

PR #445, opened 2019-12-24
Adding script for detecting device_id dynamically and appending/chang…

pareshgp:
…ing to .conf files. Removed extra newlines from end of .conf as if use this utility will add at end of file


rvs/conf/deviceid.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+# To use this utlity call it from testcript with below example
+# Example:  ./conf/deviceid.sh "conf/gpup_*.conf"
+# Above command will add device_id : XXXX in all gpup_*.conf files
+ 
+conf_to_mod=(echo $1)
```
rakeshsv: instead of $1 , can you name the variable , assign $1 to a variable

rvs/conf/deviceid.sh
```diff
@@ -0,0 +1,41 @@
...
+    sed -i '/'"${OLD_LINE_PATTERN}"'/{s/.*/'"${NEW}"'/;h};${x;/./{x;q100};x}' "${FILE}"
+    if [[ $? -ne 100 ]] && [[ ${NEW_LINE} != '' ]]
+    then
+        echo "${NEW_LINE}" >> "${FILE}"
+    fi
+}
+
+gpu_dev_id_list=()
+device_id=0
+for dir in $(ls -d /sys/class/kfd/kfd/topology/nodes/*)
```
rakeshsv: can you add comment on whats going on this loop?

rvs/conf/deviceid.sh
```diff
@@ -0,0 +1,41 @@
...
+    for j in "${gpu_dev_id_list[@]}"
+    do
+        #echo "j $j"
+        if [ '0' != '$j' ]; then    
+            d_i=$j
+        fi
+    #echo "d_i $d_i"
+    done
+done
+
```
rakeshsv: Is this where you change all the conf files?

rvs/conf/deviceid.sh
```diff
@@ -0,0 +1,41 @@
...
+for dev_id_line in $1; do
+    configLine  "device_id" "  device_id : $d_i" $dev_id_line
```
rakeshsv: whats the logic? If device ID is present then change the device IDs? 

Why are you maintaining an array of device IDs, it should be just one device ID isn't it?

rvs/conf/deviceid.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
```
rakeshsv: When this .sh file will be called?

pareshgp: This file is called from testscripts/*.sh files. So after build if the user calls any file under testscripts which is copied to build/bin ex gst.new.sh/pesm.new.sh then this .sh file get called.

rvs/conf/deviceid.sh
pareshgp: Yes. it is one device id. I maintained just to check if there are no 2 different devices. At last, assigning it to one device id and used same to modify the config file

rvs/conf/deviceid.sh
pareshgp: No. conf file will get changed when configLine function gets called and the below line will change config file.

    if [[ $? -ne 100 ]] && [[ ${NEW_LINE} != '' ]]
    then
        echo "${NEW_LINE}" >> "${FILE}"
    fi

rvs/conf/deviceid.sh
pareshgp: $1 is used to get command line arguments(passed .conf files to .sh). In case of argument gst_*.conf to check for all conf $1 assignment to a variable is converting for the first file from an array. So right now keeping it $1 and will modify once I will make it working with variable.

pareshgp:
Changed Base branch from develop to rocm-3.2. Resolved merge issue. Check for gst_worker.cpp and gst_1.conf

PR #448, opened 2020-02-16
Updating the Googletest revision to 1.10.0

pruthvistony:
Signed-off-by: Pruthvi Madugundu <mpruthvi@gmail.com>


rakeshsv:
Fix for the google test code which is causing the build to break

PR #451, opened 2020-03-04
Fixed Gflops calculation and allowed blass gemm to successfully complete

pareshgp:


pareshgp:
Added Single .cong changes for GST PEBB and PQT with 451 pull request.

PR #474, opened 2021-01-11
Remove nodes that are not connected from test

manoj-freyr:
If nodes have no connection in either direction, it makes no sense
to add to test array and run pebb as hsa internally errors out anyway

Signed-off-by: Manoj S K <Manoj.SK@amd.com>


rakeshsv (commented):
Why there wouldnt be a PCIe connection to the system? If there is none then there no GPU detected. Pebb is more of host to device and device to host bandwidth

manoj-freyr:
Defensive checks to prevent in case PCIe connection is not there between nodes.

PR #485, opened 2021-07-29
Iet smichanges

manoj-freyr:
Changes to align rocm-smi and hip enumerations

### DIFF
--- a/iet.so/include/action.h
+++ b/iet.so/include/action.h
@@ -119,7 +119,7 @@ class iet_action: public rvs::actionbase {
     //! selected for EDPp test
     std::vector<gpu_hwmon_info> edpp_gpus;
 
-
+    void hip_to_smi_indices();
     bool get_all_iet_config_keys(void);
     /**
     * @brief reads all common configuration keys from
--- a/iet.so/include/action.h
+++ b/iet.so/include/action.h
@@ -118,7 +118,7 @@ class iet_action: public rvs::actionbase {
     //! list of GPUs (along with some identification data) which are
     //! selected for EDPp test
     std::vector<gpu_hwmon_info> edpp_gpus;
-
+    std::map<int, int> hip_to_smi_idxs;
     void hip_to_smi_indices();
     bool get_all_iet_config_keys(void);
     /**
--- a/iet.so/include/iet_worker.h
+++ b/iet.so/include/iet_worker.h
@@ -62,7 +62,12 @@ class IETWorker : public rvs::ThreadBase {
     }
     //! returns the GPU index
     int get_gpu_device_index(void) { return gpu_device_index; }
-
+    //! sets the GPU smi index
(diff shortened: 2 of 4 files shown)


iet.so/include/action.h
```diff
@@ -118,8 +118,8 @@ class iet_action: public rvs::actionbase {
     //! list of GPUs (along with some identification data) which are
     //! selected for EDPp test
     std::vector<gpu_hwmon_info> edpp_gpus;
-
-
+    std::map<int, int> hip_to_smi_idxs;
```
ashutom, line 121: I am not sure of the norms followed in RVS still,
for , hip_to_smi_idxs;
we can follow Hungarian notation of 'm'  to indicat its member of that class 
( never confuse type with kind in Hungarian : https://www.joelonsoftware.com/2005/05/11/making-wrong-code-look-wrong/)

iet.so/include/iet_worker.h
```diff
@@ -62,7 +62,12 @@ class IETWorker : public rvs::ThreadBase {
     }
     //! returns the GPU index
     int get_gpu_device_index(void) { return gpu_device_index; }
-
+    //! sets the GPU smi index
+    void set_smi_device_index(int _smi_device_index) {
```
ashutom, line 66: void set_smi_device_index(int _smi_device_index)   should always be

void set_smi_device_index(const int&  _smi_device_index)

iet.so/include/iet_worker.h
```diff
@@ -208,14 +213,15 @@ class IETWorker : public rvs::ThreadBase {
...
-
  protected:
     std::unique_ptr<rvs_blas> gpu_blas;
 
     //! name of the action
     std::string action_name;
     //! index of the GPU (as reported by HIP API) that will run the EDPp test
     int gpu_device_index;
+    //! index of GPU (in view of smi lib) which is sometimes different to above index
+    int smi_device_index;
```
ashutom, line 224: Same comment as before

iet.so/include/iet_worker.h
manoj-freyr: for simpler and native data types like int, char , I dont think pass by const ref will differ much from pass by value

iet.so/include/iet_worker.h
manoj-freyr: agreed, but kept naming convention consistent with the existing fields

ashutom (commented):
Check individual comments

iet.so/include/action.h
manoj-freyr: agreed, but kept naming convention consistent with the existing fields

iet.so/include/iet_worker.h
ashutom: agreed,  but like 'winning is a habit' so is 'good coding ways' should also be :)
you may leave the code currently as it is

PR #487, opened 2021-08-17
Updated rocm-smi-lib package name in README

pramenku:
rocm-smi-lib64 is not more valid. It has been renamed to rocm-smi-lib


manoj-freyr (approved):
LGTM

PR #488, opened 2021-08-17
Running changes

manoj-freyr:
1. Fixing typos
2. Making description field in accordance with https://www.debian.org/doc/debian-policy/ch-controlfields.html#description 
3. Add shebang to specify python3 so as not to fail builds in cases where python symlink is not created.


CMakeLists.txt
```diff
@@ -108,7 +108,8 @@ get_version ( ${RVS_VERSION} )
 # set(CMAKE_PACKAGING_INSTALL_PREFIX "/opt/rocm")
 set(CPACK_PACKAGE_NAME "rocm-validation-suite")
 set(CPACK_PACKAGE_DESCRIPTION "ROCm Validation Suite")
-set(CPACK_PACKAGE_DESCRIPTION_SUMMARY "The ROCm Validation Suite – The ROCm Validation Suite is a system administrator and cluster manager's tool for detecting and troubleshooting common problems affecting AMD GPUs running in high performance computing environments, enabled using the ROCm software stack on a compatible platform.")
+set(CPACK_PACKAGE_DESCRIPTION_SUMMARY "The ROCm Validation Suite validates the AMD platform using ROCM Runtime.
+ The ROCm Validation Suite is a system administrator and cluster manager's tool for detecting and troubleshooting common problems affecting AMD GPUs running in high performance computing environments, enabled using the ROCm software stack on a compatible platform.")
```
ashutom, line 112: using ROCM Runtime  ==> Should be ROCm , as you have done with the whole sentence

Also, kindly change " The ROCm Validation Suite – "  to  "ROCm Validation Suite –" ,  since we are at 'this' package, In my view "the" should not be used.

regression/run_regression.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
ashutom, line 1: Is it not possible that you use update alternatives and then change the python to whatever is required instead of changing 10 files for that?

ashutom (commented):
Is it not possible that you use update alternatives and then change the python to whatever is required instead of changing 10 files for that?

regression/run_regression.py
manoj-freyr: Instead of adding one more script file and keeping the shebang line as such will not give me any minimization of code as the script will still have the shebang.
Also, since no python2 specific code , I guess this is simpler way to proceed.

CMakeLists.txt
manoj-freyr: Done

ashutom (approved):
Since there is backward compatibility issue with python2 , we can go ahead and modify the shebang in all the files.

Looks good to me now.

PR #492, opened 2021-09-02
Logging to identify MCM GPUs.

jkottiku:
SWDEV-291318 : [MI200] RVS IET module does not report power on one of the cards

This additional logging will help to know if the GPU die/chiplet is in Multi-Chip Module (MCM) GPU.


frepaul (approved):
LGTM

PR #493, opened 2021-09-04
Fix typo

Umio-Yasuno:
I've also created [a branch to simplify the boolean determination.](https://github.com/Umio-Yasuno/ROCmValidationSuite/tree/bool)
`val == true` --> `val`
But since this is my first pull request, I'd like to just fix the typo.


jkottiku (approved):
Looks good

PR #494, opened 2021-09-06
Cleanup and enable CMAKE_PREFIX_PATH

manoj-freyr:
CMAKE_PREFIX_PATH to append ROCM_PATH so that find_package() works


ashutom (approved):
LGTM

PR #497, opened 2021-09-16
Add compiler option "-O0" when ASAN is enabled.

jkottiku:
SWDEV-303253 : [ROCmValidationSuite] Address Sanitizer not detecting memory leaks

Due to compiler optimizations, Address Sanitizer (ASAN) was not detecting memory leaks. 
When ASAN build is enabled, set "-O0" no optimization option for compiler.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -427,7 +427,7 @@ set(RVS_BINTEST_FOLDER ${CMAKE_BINARY_DIR}/bintest)
 
 # Check if Address Sanitizer enabled.
 if(BUILD_ADDRESS_SANITIZER)
-  set(ASAN_CXX_FLAGS "-fsanitize=address  -shared-libasan")
+  set(ASAN_CXX_FLAGS "-fsanitize=address  -shared-libasan -O0")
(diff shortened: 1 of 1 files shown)


manoj-freyr (approved):
Looks good.

b-sumner:
This will slow down the execution even more, but otherwise I have no objection.

jkottiku:
> This will slow down the execution even more, but otherwise I have no objection.

Yes, thats correct. But ASAN fails to detect memory leaks with compiler optimization enabled. Hence, added "-O0" to ensure all leaks are detected properly.

b-sumner:
That may not be true.  With optimization enabled, there may be no memory leak.

jkottiku:
> That may not be true. With optimization enabled, there may be no memory leak.

Understood. Compiler might take care of it. 
Hope it is OK to have -O0 to explicitly find leaks instead of depend on compiler optimization ? Please let me know.

b-sumner:
Yes, I have no problem with -O0 if the runtime is bearable.

jkottiku:
> Yes, I have no problem with -O0 if the runtime is bearable.

Sure. Thanks a lot for your valuable inputs @b-sumner

PR #501, opened 2021-09-27
Remove unused local variables.

jkottiku:
SWDEV-304055
[ROCmValidationSuite] [Compiler Warnings] Unused variables "-Wunused-variable"

Remove unused local variables in functions to avoid compiler warnings.

This reverts commit 82e5bb53d76247b3e78a25c2eb11085cf80f8631.

### DIFF
--- a/iet.so/src/action.cpp
+++ b/iet.so/src/action.cpp
@@ -375,20 +375,22 @@ bool iet_action::get_all_common_config_keys(void) {
  * 
  */
 
-void iet_action::hip_to_smi_indices(){
+void iet_action::hip_to_smi_indices(void) {
     int hip_num_gpu_devices;
     hipGetDeviceCount(&hip_num_gpu_devices);
     // map this to smi as only these are visible
     uint32_t smi_num_devices;
     uint64_t val_ui64;
     std::map<uint64_t, int> smi_map;
+
     rsmi_status_t err = rsmi_num_monitor_devices(&smi_num_devices);
     if( err == RSMI_STATUS_SUCCESS){
-	for(auto i = 0; i < smi_num_devices; ++i){
-	    err = rsmi_dev_pci_id_get(i, &val_ui64);
-	    smi_map.insert({val_ui64, i});
-	}
+        for(auto i = 0; i < smi_num_devices; ++i){
+            err = rsmi_dev_pci_id_get(i, &val_ui64);
+            smi_map.insert({val_ui64, i});
+        }
     }
+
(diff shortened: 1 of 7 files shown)


iet.so/src/action.cpp
```diff
@@ -375,20 +375,22 @@ bool iet_action::get_all_common_config_keys(void) {
  * 
  */
 
-void iet_action::hip_to_smi_indices(){
+void iet_action::hip_to_smi_indices(void) {
```
manoj-freyr, line 378: C++ doesnt need void in function parameter. Empty parameter list means funtion expects nothing. In C, this makes sense as empty means more than 0 args

iet.so/src/action.cpp
```diff
@@ -416,34 +418,36 @@ bool iet_action::do_edp_test(map<int, uint16_t> iet_gpus_device_index) {
...
     int hip_num_gpu_devices;
+    unsigned int i = 0;
+    map<int, uint16_t>::iterator it;
+    uint32_t smi_num_devices;
+
     hipGetDeviceCount(&hip_num_gpu_devices);
 
     vector<IETWorker> workers(iet_gpus_device_index.size());
     for (;;) {
-        unsigned int i = 0;
```
manoj-freyr, line 423: typically better to define variables as late as possible. Avoids unnecessary allocations in case(if later) exceptions are thrown before their usage. Though in this case not much construction or destruction takes place other than iterator, it would be better to declare them just before we need it and just before we have values needed to initialize it.

src/rvsliblogger.cpp
```diff
@@ -351,9 +351,8 @@ int rvs::logger::JsonStartNodeCreate(const char* Module, const char* Action) {
  *
  */
 
-int rvs::logger::JsonEndNodeCreate() {
+int rvs::logger::JsonEndNodeCreate(void) {
```
manoj-freyr, line 354: same comment as above.

iet.so/src/action.cpp
jkottiku: Thought of separating declarations statements from executable statements. Will revert this change for now as there other also following the same method.

iet.so/src/action.cpp
jkottiku: True. Just being explicit here :-)

src/rvsliblogger.cpp
jkottiku: True. Just being explicit here :-)

manoj-freyr (approved):

PR #513, opened 2021-11-18
Removal of unnecessary confs

manoj-freyr:
Removing conf files that are cluttering master branch ..


jkottiku (commented):
1. QA automation testing is not using any of these confs. Right ?
2. In module conf, we are covering all basic use cases ?

manoj-freyr:
> 1. QA automation testing is not using any of these confs. Right ?
> 2. In module conf, we are covering all basic use cases ?

Yes and Yes.
Also if we resuscitate other modules, we can fetch back from develop branch.
This clean up avoids unnecessary clutter and also avoids confusion among users on myriad configs for each module.

jkottiku (approved):

PR #517, opened 2021-12-01
Change json logging for throughput & link type

jkottiku:
In PQT module, change json logging for throughput & link type.


pqt.so/include/action.h
```diff
@@ -48,6 +48,11 @@ using namespace std::chrono;
 
 class pqtworker;
 
+typedef enum class pqt_json_data {
```
manoj-freyr, line 51: I dont think a typedef is needed. Enum class defines a type.

manoj-freyr (approved):
LGTM

pqt.so/include/action.h
jkottiku: Yes, it's implicit. Removed typedef.

PR #519, opened 2021-12-22
Check for gst completed GEMM

manoj-freyr:
Adding a check if gemm task is  complete or not

### DIFF
--- a/gst.so/src/gst_worker.cpp
+++ b/gst.so/src/gst_worker.cpp
@@ -425,7 +425,7 @@ bool GSTWorker::do_gst_stress_test(int *error, std::string *err_description) {
 
         // run GEMM & wait for completion
         gpu_blas->run_blass_gemm(gst_ops_type);
-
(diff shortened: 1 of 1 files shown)


gst.so/src/gst_worker.cpp
```diff
@@ -425,7 +425,7 @@ bool GSTWorker::do_gst_stress_test(int *error, std::string *err_description) {
 
         // run GEMM & wait for completion
         gpu_blas->run_blass_gemm(gst_ops_type);
-
+        while (!gpu_blas->is_gemm_op_complete()) {} // confirm process is complete
```
jkottiku, line 428: "Wait for the previous gemm operations to complete" would be a better comment ? Please check.

jkottiku: Good catch. We have done the same in IET too.

manoj-freyr: yes

jkottiku (approved):

PR #525, opened 2022-02-16
recognize arch distro

acxz:
This PR allows ArchLinux to be recognized as a distro and ensures that `command_string` is defined in that case.

Fixes: #526


manoj-freyr (changes requested):
Changes has to go to master branch which is the ToT.
and RCQT is being changed and refactored because of which this change becomes a moot point. Kindly check our changes where we will commit RCQT refactored and redesigned changes.

acxz:
> Changes has to go to master branch which is the ToT.

Got it, will change base branch to master

> and RCQT is being changed and refactored because of which this change becomes a moot point. Kindly check our changes where we will commit RCQT refactored and redesigned changes

I would kindly suggest that instead of waiting for a feature to get merged in and having broken functionality in the meanwhile, it would be better to have this merged.

But I'm happy that the new feature will fix the root issue at hand! Can't wait till it lands. Till then we'll be using this patch over at Archlinux

>  Also, ROCM(and RVS) currently is not having full fldged support to arch.

That is totally understandable, however, this little fix will at least allow us to try to build with arch, instead of artificially breaking the build based on OS type.

Thank you for the review.

PR #529, opened 2022-03-03
Change value of ramp interval

jkottiku:
[SWDEV-325774] - 
Set ramp interval to 5 mins in 
1. gst_stress_12_hrs.conf
2. gst_stress_6_hrs.conf
3. gst_stress_3_hrs.conf
which will result in actual test duration almost equal to duration.


manoj-freyr (approved):
LGTM. This is an arbitrary interval for ramping and we can tweak it as the regressions proceed

PR #531, opened 2022-03-16
 File reorganization as per the new defined standard.

nunnikri:
The binary will be in /opt/rocm-xxx/bin
The private binaries like internal libs will be /opt/rocm-xxx/lib/rvs
The shared conf, doc/man and test scripts in /opt/rocm-xxx/share


jkottiku (approved):
Changes looks good.
Hope we have also run some RVS test ?

manoj-freyr (approved):
LGTM

manoj-freyr (approved):
LGTM

PR #532, opened 2022-03-29
Mem refactor

manoj-freyr:
Make configurable tests

### DIFF
--- a/mem.so/src/rvs_memtest.cpp
+++ b/mem.so/src/rvs_memtest.cpp
@@ -678,11 +678,11 @@ void test3(char* ptr, unsigned int tot_num_blocks)
     err = move_inv_test(ptr, tot_num_blocks, p1, p2);
 
     if(!err) {
-         msg = "[" + memdata.action_name + "] " + MODULE_NAME + " " + "Test 4 : PASS";
+         msg = "[" + memdata.action_name + "] " + MODULE_NAME + " " + "Moving inversions successful";
          rvs::lp::Log(msg, rvs::loginfo);
     }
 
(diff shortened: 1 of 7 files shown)


mem.so/src/rvs_memtest.cpp
```diff
@@ -114,53 +99,59 @@ unsigned int error_checking(const std::string& pmsg, unsigned int blockidx)
...
-    }
+    hipMemset(ptCntOfError, 0, sizeof(unsigned int));
+    hipMemset((void*)&ptFailedAdress[0], 0, sizeof(unsigned long)*MAX_ERR_RECORD_COUNT);;
+    hipMemset((void*)&ptExpectedValue[0], 0, sizeof(unsigned long)*MAX_ERR_RECORD_COUNT);;
+	  hipMemset((void*)&ptCurrentValue[0], 0, sizeof(unsigned long)*MAX_ERR_RECORD_COUNT);;
+
+    hipDeviceReset();
+    exit(ERR_BAD_STATE);
 
-    return numOfErrors;
```
jkottiku, line 163: **return** statement missing.

jkottiku (changes requested):

mem.so/src/rvs_memtest.cpp
manoj-freyr: Line 107 actually returns if no errors and in case errors we are calling exit(ERR_BAD_STATE). This is same as earlier case but re-organized code for clarity.

jkottiku (approved):
Looks good.

PR #537, opened 2022-04-05
Pqt hang

manoj-freyr:
The way we call hsa APIs for copy, signal_wait etc changed and removed unnecessary iterations on non peer GPUs.
GPU requested access for destination memory pools . Cleaned up logic.

### DIFF
--- a/pqt.so/src/action.cpp
+++ b/pqt.so/src/action.cpp
@@ -510,35 +510,25 @@ int pqt_action::create_threads() {
           pqtworker* p = nullptr;
 
           transfer_ix += 1;
-          if (b2b_block_size > 0 && property_parallel) {
-            RVSTRACE_
-            pqtworker* pb2b = new pqtworker;
-            if (pb2b == nullptr) {
-              RVSTRACE_
-              msg = "internal error";
-              rvs::lp::Err(msg, MODULE_NAME_CAPS, action_name);
-              return -1;
-            }
-            pb2b->initialize(srcnode, dstnode, prop_bidirectional);//,
-                             //b2b_block_size);
-            p = pb2b;
 
-          } else {
+          p = new pqtworker;
+          if (p == nullptr) {
             RVSTRACE_
-            p = new pqtworker;
-            if (p == nullptr) {
-              RVSTRACE_
-              msg = "internal error";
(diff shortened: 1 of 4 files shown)


pqt.so/src/action.cpp
```diff
@@ -402,7 +402,7 @@ int pqt_action::create_threads() {
   gpu_get_all_gpu_id(&gpu_id);
   gpu_get_all_device_id(&gpu_device_id);
 
-  for (size_t i = 0; i < gpu_id.size(); i++) {    // all possible sources
+  for (size_t i = 0; i < gpu_id.size()-1; i++) {    // all possible sources
```
jkottiku, line 405: Since its <, for full iteration we need **gpu_id.size()**. Right ?

jkottiku (commented):
Let's sync up on the changes.

pqt.so/src/action.cpp
manoj-freyr: Yeah this is src. I mean it should have atleast one destination to do transfer right? So i will be from 0 to n-1 and j will be from i+1 to n, thus we can cover all. i reaching n will result in no destinations

pqt.so/src/action.cpp
```diff
@@ -508,35 +510,25 @@ int pqt_action::create_threads() {
...
+            rvs::lp::Err(msg, MODULE_NAME_CAPS, action_name);
+            return -1;
           }
+          p->initialize(srcnode, dstnode, prop_bidirectional);
           RVSTRACE_
           p->set_name(action_name);
           p->set_stop_name(action_name);
           p->set_transfer_ix(transfer_ix);
-          p->set_block_sizes(block_size);
+          if (b2b_block_size > 0 && property_parallel) {
```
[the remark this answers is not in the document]
manoj-freyr: Agreed. Practically b2b_block_size is not used and its functionality can be subsumed in regular worker object with block_size variable. Will remove it.

jkottiku:
Please correct function header of function rvs::hsa::SendP2PTraffic(). Rest looks good.

jkottiku (approved):
Please correct function header of function rvs::hsa::SendP2PTraffic(). Rest looks good.

PR #539, opened 2022-04-19
List all the ROCm packages and its version installed in system.

jkottiku:
rcqt: list all the ROCm packages and its version installed in the system.

### DIFF
--- a/rcqt.so/include/debPackageInfo.h
+++ b/rcqt.so/include/debPackageInfo.h
@@ -11,6 +11,13 @@ class DebPackageInfo : public PackageInfo{
           cmd[0],
           cmd[1],
           cmd[2]){}
+
+    DebPackageInfo(const std::string& pkgmgr,
+        const std::vector <std::string> cmd):PackageInfo(pkgmgr,
+          cmd[0],
+          cmd[1],
+          cmd[2]){}
+
     bool readMetaPackageInfo(std::string ss) override;
     std::pair<std::string, std::string> getNameVers(std::string word);
     void readDepLine(const std::string& depLine);
--- a/rcqt.so/include/debPackageInfo.h
+++ b/rcqt.so/include/debPackageInfo.h
@@ -5,18 +5,13 @@
 class DebPackageInfo : public PackageInfo{
   public:
     DebPackageInfo(const std::string& pkgname,
-        const std::string& pkgmgr,
         const std::vector <std::string> cmd):PackageInfo(pkgname,
-          pkgmgr,
-          cmd[0],
-          cmd[1],
-          cmd[2]){}
+          std::string("dpkg"),
+          cmd){}
 
-    DebPackageInfo(const std::string& pkgmgr,
-        const std::vector <std::string> cmd):PackageInfo(pkgmgr,
-          cmd[0],
-          cmd[1],
-          cmd[2]){}
+    DebPackageInfo(const std::vector <std::string> cmd):PackageInfo(
+          std::string("dpkg"),
+          cmd){}
 
     bool readMetaPackageInfo(std::string ss) override;
     std::pair<std::string, std::string> getNameVers(std::string word);
--- a/rcqt.so/include/metaPackageInfo.h
+++ b/rcqt.so/include/metaPackageInfo.h
@@ -8,10 +8,15 @@ class PackageInfo {
 
 public:
   PackageInfo(const std::string& pkgname, const std::string& pkgmgr,
-      const std::string& cmd1, const std::string& cmd2 = "", const std::string& cmd3 = "") :
+      const std::string& cmd1, const std::string& cmd2, const std::string& cmd3) :
     m_pkgname{pkgname}, m_pkgmgrname{pkgmgr}, m_dependcmd1name{cmd1}, m_dependcmd2name{cmd2}, m_infocmdname{cmd3} {
       if(!pkgname.empty())
         m_filename = pfilename(pkgname);	
+    }
+
+  PackageInfo(const std::string& pkgmgr,
+      const std::string& cmd1, const std::string& cmd2, const std::string& cmd3) :
+    m_pkgmgrname{pkgmgr}, m_dependcmd1name{cmd1}, m_dependcmd2name{cmd2}, m_infocmdname{cmd3} {
     }
 
 	virtual ~PackageInfo(){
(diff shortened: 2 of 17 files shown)


rcqt.so/include/debPackageInfo.h
```diff
@@ -11,6 +11,13 @@ class DebPackageInfo : public PackageInfo{
           cmd[0],
           cmd[1],
           cmd[2]){}
+
+    DebPackageInfo(const std::string& pkgmgr,
+        const std::vector <std::string> cmd):PackageInfo(pkgmgr,
+          cmd[0],
```
manoj-freyr, line 17: this again somehow binds to the number. Like a vector size being 3 or more? Why not pass the vector down the hierarchy to base class ctor also and let it unpack it? that would be more extensible and you limit the number of parameters to ctor too.

rcqt.so/include/metaPackageInfo.h
```diff
@@ -8,10 +8,15 @@ class PackageInfo {
 
 public:
   PackageInfo(const std::string& pkgname, const std::string& pkgmgr,
-      const std::string& cmd1, const std::string& cmd2 = "", const std::string& cmd3 = "") :
+      const std::string& cmd1, const std::string& cmd2, const std::string& cmd3) :
```
manoj-freyr, line 11: same comment as above. can we take inputs as collection itself than cmd1, cmd2, cmd3? It is ambiguous to reader as what are cmd1 etc

rcqt.so/include/packageHandler.h
```diff
@@ -30,11 +30,20 @@ class PackageHandler{
...
+	}
+  
+  void setPackageList(const std::vector<std::string>& pkglist) {
+    m_pkglist = pkglist;
+	}
+	const std::vector<std::string>& getPackageList() const {
+		return m_pkglist;
+	}
+  void listPackageVersion();
 	virtual ~PackageHandler(){}
```
manoj-freyr, line 42: indentation

rcqt.so/include/rpmPackageInfo.h
```diff
@@ -12,6 +13,13 @@ class RpmPackageInfo : public PackageInfo {
           cmd[0],
           cmd[1],
           cmd[2]){}
+
+    RpmPackageInfo(const std::string& pkgmgr,
+        const std::vector <std::string> cmd):PackageInfo(pkgmgr,
```
manoj-freyr, line 18: same comment as above to passing as collection

rcqt.so/src/action.cpp
```diff
@@ -55,6 +55,7 @@
 #define JSON_CREATE_NODE_ERROR "JSON cannot create node"
 #define JSON_PKGCHK_NODE_NAME "pkgchk"
 #define PACKAGE "package"
+#define PACKAGELIST "packagelist"
```
manoj-freyr, line 58: just another general comment , or a guideline for us to follow, to use static const variables instead of preprocessor directives , atleast going forward.

rcqt.so/src/action.cpp
```diff
@@ -106,6 +107,7 @@ rcqt_action::~rcqt_action() {
 int rcqt_action::run() {
   string msg;
   bool pkgchk_bool = false;
+  bool pkglist_bool = false;
```
manoj-freyr, line 110: Can we do away with having  data type in variable name? i.e., "bool" in name

rcqt.so/src/packageHandler.cpp
```diff
@@ -66,3 +66,35 @@ void PackageHandler::validatePackages(){
...
+		return;
+	}
+	int totalPackages = 0, missingPackages = 0, installedPackages = 0;
+
+	for (const auto& pkgname: pkglist){
+
+		++totalPackages;
+
+		auto installedversion = getInstalledVersion(pkgname);
+    if(!installedversion.empty()){
```
manoj-freyr, line 83: indentations

rcqt.so/src/packageHandler.cpp
```diff
@@ -66,3 +66,35 @@ void PackageHandler::validatePackages(){
...
+			++installedPackages;
+			std::cout << "Package " << pkgname << " installed version is " << 
+					installedversion << std::endl;
+    }
+    else {
+      ++missingPackages;
+      std::cout << "Package " << pkgname << " not installed " <<
+        std::endl;
```
manoj-freyr, line 91: indentations

rcqt.so/src/packageHandlerDeb.cpp
```diff
@@ -18,6 +18,13 @@ PackageHandlerDeb::PackageHandlerDeb(std::string pkgname): PackageHandler{}{
   m_manifest = metaInfo->getFileName();
 }
 
+PackageHandlerDeb::PackageHandlerDeb(): PackageHandler{}{
+  
+  const std::vector<std::string> cmd {std::string("--status"), std::string(""), std::string("--status")};
+
+	metaInfo.reset(new DebPackageInfo(std::string("dpkg"), cmd));
```
manoj-freyr, line 25: can depPackageInfo or RPMPackageInfo etc have "dpkg" or "rpm" in their class definition itself? instead of us passing it down? They encapsulate the state specific to that package right? what do you think? We have in ctor, DebPackageInfo which is a Deb specific class and why do we need to still pass dpkg in it as dpkg is specific to Deb class.

manoj-freyr (changes requested):

rcqt.so/include/packageHandler.h
jkottiku: done

rcqt.so/src/packageHandler.cpp
jkottiku: Done

rcqt.so/src/packageHandler.cpp
jkottiku: Done

rcqt.so/src/action.cpp
jkottiku: Yes

rcqt.so/src/action.cpp
jkottiku: Currently, we have preprocessor directives in RVS for property variables. Need to see if static const variables really does any value add.

jkottiku:
@manoj-freyr Review comments fixed. Please have a look.

manoj-freyr (approved):
LGTM

PR #541, opened 2022-05-31
Allow building on debian

superzerg:
I was able to build rvm on debian with this small change.


acxz:
@jkottiku can this be closed on the same reasoning as https://github.com/ROCm-Developer-Tools/ROCmValidationSuite/pull/534#issuecomment-1158327573 ?

superzerg:
Would you prefer to have some warning like `WARNING: debian is NOT officially supported...` displayed in case of debian or arch ?

acxz:
@jkottiku 

> can this be closed

I'm actually asking you to close this

PR #544, opened 2022-06-17
Update lib path as per reorg. changes.

jkottiku:
SWDEV-342283

In the new reorg., RVS libs got moved to lib64/rvs from lib/rvs folder which caused this issue.
As RVS does dynamic loading of shared libs from the code, path change will be done in the RVS code to fix this issue.

### DIFF
--- a/rvs/src/rvsmodule.cpp
+++ b/rvs/src/rvsmodule.cpp
@@ -165,26 +165,35 @@ rvs::module* rvs::module::find_create_module(const char* name) {
       libpath += "/";
     } else {
       rvs::options::has_option("pwd", &libpath); // has ending forward slash too
-      libpath += "../lib/rvs/";
+      libpath += "../lib64/rvs/";
     }
     string sofullname(libpath + it->second);
     void* psolib = dlopen(sofullname.c_str(), RTLD_NOW);
-    // error?
     if (!psolib) {
-      //Search libraries in current path for backward compatibility
+      /* Backward compatibility support */
+
+      /* Search libraries in lib/rvs folder (pre-reorg path) */ 
       rvs::options::has_option("pwd", &libpath); // has ending forward slash too
+      libpath += "../lib/rvs/";
       string sofullname(libpath + it->second);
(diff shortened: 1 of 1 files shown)


rvs/src/rvsmodule.cpp
```diff
@@ -165,26 +165,35 @@ rvs::module* rvs::module::find_create_module(const char* name) {
       libpath += "/";
     } else {
       rvs::options::has_option("pwd", &libpath); // has ending forward slash too
-      libpath += "../lib/rvs/";
+      libpath += "../lib64/rvs/";
```
nunnikri, line 168: We should change this to use CMAKE_INSTALL_LIBDIR. So that we may not need 2 path check, might need only 1 check here.

nunnikri (changes requested):

rvs/src/rvsmodule.cpp
nunnikri: Looks like the default value setting for CMAKE_INSTALL_LIBDIR is missing in rvs, that seems to have caused the move from lib to lib64. We should add that first and change here also.

jkottiku: @nunnikri Ok. We can make that change. But this implementation will work for lib, lib64 and current folder so we will not have to add anything here unless we have a new path of libs.

nunnikri: Since we moved to GNU_INSTALL_DIR, it will change based on distros. Major ones are as below.
Ubuntu : lib
RHEL/CENTOS/SLES: lib64
DEBIAN: lib/<multiarch-tuple>

Other distros it might be even different. Since we are not building in Debian in our CI, we dont get other folder. Yes this will work for now to resolve the ticket.  But its better to use CMAKE_INSTALL_LIBDIR, instead of adding cases for each and every cases.

jkottiku: Yes, CMAKE_INSTALL_LIBDIR have to be done for lib path standardization across currently supported distros.

PR #555, opened 2022-08-22
Corrected conf files and related fixes

jkottiku:
1. Corrected conf files - missing proper values
2. gst module - hot calls rollover condition handled
3. peqt conf - dos2unix converted

### DIFF
--- a/gst.so/src/gst_worker.cpp
+++ b/gst.so/src/gst_worker.cpp
@@ -472,10 +472,6 @@ bool GSTWorker::do_gst_stress_test(int *error, std::string *err_description) {
                    std::to_string(gpu_id) + " " + GST_START_MSG + " " +
                    " Executing hot calls loop :" + std::to_string(gst_hot_calls); 
             rvs::lp::Log(msg, rvs::logtrace);
-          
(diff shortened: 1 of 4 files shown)


manoj-freyr (approved):
LGTM

gst.so/src/gst_worker.cpp
```diff
@@ -472,8 +472,10 @@ bool GSTWorker::do_gst_stress_test(int *error, std::string *err_description) {
                    std::to_string(gpu_id) + " " + GST_START_MSG + " " +
                    " Executing hot calls loop :" + std::to_string(gst_hot_calls); 
             rvs::lp::Log(msg, rvs::logtrace);
-           
-            gst_hot_calls--; 
+          
+            if(0 != gst_hot_calls) {
```
manoj-freyr, line 476: Is this necessary? Because type of gst_hot_calls is unsigned int which will never be negative. Derfault value is 0 and when we check in the "if" case, we check if it is 0(with !gst_hot_calls). It enters else case only when it is not zero, so this line is redundant?

manoj-freyr (approved):
Looks good

PR #557, opened 2022-10-07
make regex sensible instead of fixing numbers

manoj-freyr:
make sure peqt doesnt depend on deviceid.sh file and no edit after installs. Fix a bug in action.cpp for correct function in pci_caps
deviceid is a field to be fetched from pci in code, removing it from first level confs as it depended on deviceid.sh file. In case filtering needed, provided a default of 0 which can be replaced with appropriate device id by end users,.

Pattern matching in peqt should be "patterns" instead of some hardcoded values which may not exactly match depending on GPU capabilities. Updated the same to reflect


jkottiku (changes requested):

src/pci_caps.cpp
```diff
@@ -317,7 +317,7 @@ void get_kernel_driver(struct pci_dev *dev, char *buff) {
     //not supported any longer
     //the more simpler way is sudo dkms status
     snprintf(buff, PCI_CAP_DATA_MAX_BUF_SIZE, "%s", PCI_CAP_NOT_SUPPORTED);
-    return;
+    //return;
```
jkottiku, line 320: Better we remove this commented line.

manoj-freyr: done

jkottiku (approved):

PR #558, opened 2022-10-13
SWDEV-345870 - Correct HSA path for new directory layout

raramakr:
With file reorganization, HSA installed in /opt/rocm-ver Use the installed path rather than using backward compatibility path /opt/rocm-ver/hsa

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -463,6 +463,15 @@ else()
 endif()
 set(ROCM_SMI_LIB "rocm_smi64" CACHE STRING "rocm_smi library name")
 
+# Determine HSA_PATH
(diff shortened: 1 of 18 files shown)


jkottiku:
Change looks good.
But would this change work with RoCM 5.3 release ? i.e HSA install path already changed in RoCM 5.3 ? Please confirm.

raramakr:
> Change looks good. But would this change work with RoCM 5.3 release ? i.e HSA install path already changed in RoCM 5.3 ? Please confirm.

5.3 release do have file reorganization changes for HSA. This will work with 5.3

jkottiku (approved):

PR #560, opened 2022-10-20
Add amd in place of rocm tools, in licenses

manoj-freyr:


jkottiku (approved):

jkottiku:
Changes looks good. On another note, should we have license header in conf. files and test scripts also ?

jkottiku (approved):

PR #561, opened 2022-10-21
license in scripts

manoj-freyr:


rvs_os_helper.sh
```diff
@@ -1,2 +1,27 @@
...
+# # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# # SOFTWARE.
+# #
+# ###############################################################################
+
 #!/bin/bash
-cat /etc/os-release|grep ^ID=|sed 's/ID=//;s/"//g'
\ No newline at end of file
+cat /etc/os-release|grep ^ID=|sed 's/ID=//;s/"//g'
```
jkottiku, line 27: This should work in all OSs Right ?

jkottiku (approved):

rvs_os_helper.sh
manoj-freyr: Yes this does, it is same since repackaging changes so no new addition is been made in this CR other than adding license banner to it

PR #570, opened 2022-12-02
Initialize size variables for blas operations

jkottiku:
1. Initialize size variables for blas operations
2. Allocate size_d memory in gpu only for hgemm operation


manoj-freyr (approved):
lgtm

PR #574, opened 2022-12-17
Replace the deprecated compiler option "--amdgpu-target"

jkottiku:
Replace the deprecated compiler option "--amdgpu-target" with "--offload-arch".


manoj-freyr (approved):
LGTM, from review of https://docs.amd.com/en-US/bundle/ROCm-Compiler-Reference-Guide-v5.2/page/Additional_Compiler_Options_and_Features.html

PR #582, opened 2023-02-20
MCM check not depending on device id

manoj-freyr:


src/gpu_util.cpp
```diff
@@ -275,7 +275,6 @@ void gpu_get_all_domain_id(std::vector<uint16_t>* pgpus_domain_id,
  * @brief Check if the GPU is die (chiplet) in Multi-Chip Module (MCM) GPU.
  * @param device_id GPU Device ID
  * @return true if GPU is die in MCM GPU, false if GPU is single die GPU.
- **/
 bool gpu_check_if_mcm_die (uint16_t device_id) {
```
jkottiku, line 278: This implementation could be removed. Right ?

src/gpu_util.cpp
```diff
@@ -290,6 +289,29 @@ bool gpu_check_if_mcm_die (uint16_t device_id) {
...
 }
 
+ **/
+
+/**
+ * @brief Check if the GPU is die (chiplet) in Multi-Chip Module (MCM) GPU.
+ * @param device_id GPU Device ID
+ * @return true if GPU is die in MCM GPU, false if GPU is single die GPU.
+ **/
+bool gpu_check_if_mcm_die (uint16_t device_id, int idx) {
```
jkottiku, line 299: device_id unused in the function.

src/gpu_util.cpp
```diff
@@ -296,7 +296,7 @@ bool gpu_check_if_mcm_die (uint16_t device_id) {
  * @param device_id GPU Device ID
```
jkottiku, line 278: Correct this too :-)

jkottiku (approved):

src/rvs_util.cpp
```diff
@@ -161,12 +161,11 @@ bool fetch_gpu_list(int hip_num_gpu_devices, map<int, uint16_t>& gpus_device_ind
                 amd_mcm_gpu_found = true;
             }
 	}
-	if (cur_gpu_selected) {
-		if(!mcm_check || !mcm_die ){
+	// excluding secondary die from test list, drops power reading substantially
```
jkottiku, line 164: Comment shall be something like "Add both primary and secondary GPU die to the device list" and the current comment shall be a note in the comment section.

jkottiku (approved):

PR #584, opened 2023-03-04
Nv32

manoj-freyr:
Changes to make specific confs for Navi32 and fix certain logging inconsistencies in IET reporting

### DIFF
--- a/iet.so/src/iet_worker.cpp
+++ b/iet.so/src/iet_worker.cpp
@@ -232,7 +232,7 @@ bool IETWorker::do_iet_power_stress(void) {
             std::to_string(gpu_id) + " " + " Target power is : " + " " + std::to_string(target_power);
         rvs::lp::Log(msg, rvs::logtrace);
 
-        //check whether we reached the target power
+        //update power to max if it is valid
         if(cur_power_value > 0){
             max_power = std::max(max_power, cur_power_value);// max of averages
         }
@@ -267,6 +267,7 @@ bool IETWorker::do_iet_power_stress(void) {
     // json log the avg power
     log_to_json(IET_AVERAGE_POWER_KEY, std::to_string(max_power),
             rvs::loginfo);
+    //check whether we reached the target power
     if(max_power >= target_power) {
         msg = "[" + action_name + "] " + MODULE_NAME + " " +
             std::to_string(gpu_id) + " " + " Average power met the target power :" + " " + std::to_string(max_power);
--- a/iet.so/src/iet_worker.cpp
+++ b/iet.so/src/iet_worker.cpp
@@ -242,7 +242,7 @@ bool IETWorker::do_iet_power_stress(void) {
         total_time_ms = time_diff(end_time, iet_start_time);
 
         msg = "[" + action_name + "] " + MODULE_NAME + " " +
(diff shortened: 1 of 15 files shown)


jkottiku:
Can you please provide a short description of the change in place of "Nv32" ?

src/rvs_blas.cpp
```diff
@@ -437,8 +436,8 @@ bool rvs_blas::run_blass_gemm(std::string ops_type) {
                   int sol_index = 0;
                   int flags = 10;
 
-                  alpha = blas_alpha_val;
-                  beta  = blas_beta_val;
+                  alpha.data = blas_alpha_val;
+                  beta.data  = blas_beta_val;
```
TorreZuk, line 440: This is wrong data conversion.  If you call hgemm you need to use a IEEE float16 representation.  You may need compiler flags and then just _Float16 alpha = (float)blas_alpha_value.     I add the cast as there was a bug in double to _Float16 in recent versions of the compiler.   Rocblas has a define ROCM_USE_FLOAT16 for internally not using the struct.  .data is an unsigned short so you would have serious data loss.

manoj-freyr: will fix this

iet.so/src/iet_worker.cpp
```diff
@@ -233,7 +233,7 @@ bool IETWorker::do_iet_power_stress(void) {
         rvs::lp::Log(msg, rvs::logtrace);
 
         //check whether we reached the target power
```
jkottiku, line 235: Please update the comment as well.

jkottiku (changes requested):

rvslib/CMakeLists.txt
```diff
@@ -157,6 +157,8 @@ add_library(${RVS_TARGET}rt  ${SOURCES_RT})
 
 ## define rvslib library
 add_library(${RVS_TARGET}  ${SOURCES})
+target_compile_options(${RVS_TARGET} PRIVATE -mf16c)
+target_compile_definitions(${RVS_TARGET} PRIVATE ROCM_USE_FLOAT16)
 add_dependencies(${RVS_TARGET} ${RVS_TARGET}rt)
```
TorreZuk, line 162: So  this define exposes:
#ifdef ROCM_USE_FLOAT16
typedef _Float16 rocblas_half;
#else
But using _Float16 is just as fine.

src/rvs_blas.cpp
```diff
@@ -426,8 +425,8 @@ bool rvs_blas::run_blass_gemm(std::string ops_type) {
        }
 
       if(ops_type == "hgemm") {
-                  rocblas_float alpha;
-                  rocblas_float beta;
+                  //rocblas_half alpha;
+                  //rocblas_half beta;
```
TorreZuk, line 429: These will be fine, but okay to do what you did below and just use _Float16.

TorreZuk (approved):
Very good it looks like you will have valid test data now (same requirements for both hgemm and gemm_ex for half).    Before you were getting only int truncation as the compiler converted the float to an unsigned int.

iet.so/src/iet_worker.cpp
manoj-freyr: done.

jkottiku (approved):

PR #623, opened 2023-06-14
Use CPACK_PACKAGING_INSTALL_PREFIX for prefix

manoj-freyr:
instead of non standard CMAKE_PACKAGING_INSTALL_PREFIX use CPACK_PACKAGING_INSTALL_PREFIX  as installation prefix for RVS


jkottiku (approved):
Looks good !

PR #634, opened 2023-07-26
Include suppport for mariner OS

manoj-freyr:


jkottiku:
We have following updations also as part of new OS addition in RVS
1. rcqt module
2. RVS documentation

jkottiku (approved):

PR #643, opened 2023-09-06
SWDEV-415095 - Use relative paths for RUNPATH rather than hard coded ones

raramakr:
RVS libraries are installed in /opt/rocm-ver/lib and /opt/rocm-ver/lib/rvs 
RUNPATH should be : $ORIGIN:ORIGIN/..

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,8 +69,9 @@ endif(rocblas_FOUND)
 # all find_packages relevant to this build will be in ROCM path hence appending it to CMAKE_PREFIX_PATH 
 set(ROCM_PATH "/opt/rocm" CACHE PATH "ROCM install path")
 set(CMAKE_INSTALL_PREFIX "/opt/rocm" CACHE PATH "CMAKE installation directory")
-set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}/rvs")
-set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
+# Use relative RPATH. $ORIGIN is set from build scripts.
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -24,6 +24,8 @@
 ################################################################################
 
 cmake_minimum_required ( VERSION 3.5.0 )
+set(CMAKE_SHARED_LINKER_FLAGS_INIT "-Wl,--enable-new-dtags,--rpath,$ORIGIN:$ORIGIN/.." CACHE STRING "RUNPATH for libraries")
```
jkottiku, line 27: Any reason these variables are set before the project() ?

raramakr: The  linker flags can be set before and after the project  **without using as a cache** 
But for using the same as cache variable , it need to be set before the project
https://stackoverflow.com/questions/32698580/cmake-global-linker-flag-setting-for-all-targets-in-directory

jkottiku: Ok. Also please add a comment on why these variables are required.

jkottiku: Upon build and rvs package installation, rvs bin is not find libyaml-cpp.so.0.7 path even though the lib is present.
	**libyaml-cpp.so.0.7 => not found** - Observed
	**libyaml-cpp.so.0.7 => /usr/local/lib64/libyaml-cpp.so.0.7 (0x00007f2d0b42c000)** - Expected

Before **CMAKE_INSTALL_RPATH_USE_LINK_PATH** was helping in setting this search path in rpath. In this commit, it was removed.

jkottiku (changes requested):

jkottiku (approved):
Build and verified. Looks good for now. 
Note: Might have to revisit this in CentOS if move from libyaml-cpp static lib to dynamic lib.

PR #644, opened 2023-09-08
Add support to link speeds 32.0GT/s

manoj-freyr:
For speeds higher than 16GT/s, we are getting unknown speed  in peqt logs, adding support to 32GT/s.
Constants are taken from: https://github.com/torvalds/linux/blob/master/include/uapi/linux/pci_regs.h for PCIe

### DIFF
--- a/src/pci_caps.cpp
+++ b/src/pci_caps.cpp
@@ -110,6 +110,10 @@ void get_link_cap_max_speed(struct pci_dev *dev, char *buff) {
         case 4:
             link_max_speed = "16 GT/s";
             break;
+        case 5:
+            link_max_speed = "32 GT/s";
+            break;
+
         default:
             link_max_speed = "Unknown speed";
         }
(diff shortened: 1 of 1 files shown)


src/pci_caps.cpp
```diff
@@ -172,6 +176,13 @@ void get_link_stat_cur_speed(struct pci_dev *dev, char *buff) {
             link_cur_speed = "16 GT/s";
             break;
 #endif
+
+#ifdef PCI_EXP_LNKSTA_CLS_32_0GB
```
jkottiku, line 180: Where we expect this macro to be defined ?

jkottiku (approved):

src/pci_caps.cpp
manoj-freyr: Constants are taken from: https://github.com/torvalds/linux/blob/master/include/uapi/linux/pci_regs.h for PCIe

PR #646, opened 2023-09-14
1. Updated rcqt conf file for deb & rpm package list.

jkottiku:
2. Corrected rpm package version parsing.
3. Updated rcqt conf file for meta packages.
4. Results logging.


rcqt.so/src/action.cpp
```diff
@@ -55,7 +55,7 @@
 #define JSON_CREATE_NODE_ERROR "JSON cannot create node"
 #define JSON_PKGCHK_NODE_NAME "pkgchk"
 #define PACKAGE "package"
-#define PACKAGELIST "packagelist"
+#define PACKAGELIST (getOS() == OSType::Ubuntu?"debpackagelist":"rpmpackagelist")
```
manoj-freyr, line 58: everytime PACKAGELIST appears, a call will be made to getOS() because of #define .  since this is constant to a machine, use const variable initialized to this value, which is definitely cheaper than calls to getOS()

rcqt.so/src/rpmPackageInfo.cpp
```diff
@@ -35,11 +35,17 @@ bool RpmPackageInfo::readMetaPackageInfo(std::string ss){
...
   size_t pos = std::string::npos;
   std::ofstream os;
+  size_t relPos = std::string::npos;
 
   os.open(getFileName() , std::ofstream::out | std::ofstream::app);
 
   while(std::getline(inss, line)){
 
+    /* Exclude rpmlib* packages that are not part ROCm */
+    if(0 == line.find("rpmlib")) {
```
manoj-freyr, line 45: is it guaranteed that rpmlib packages start with this pattern always? Or is comparing against npos better?

jkottiku: Yes, thats what I have seen. Those packages start with rpmlib*.

manoj-freyr (approved):
looks good

PR #647, opened 2023-09-19
navi32 changes for sgemm

manoj-freyr:
picking specific commit onto release


jkottiku (commented):
The changes are good. But aren't we merging it to mainline as well ?

jkottiku (approved):

PR #658, opened 2023-10-10
correct names of packages and fix - issue

manoj-freyr:
1.rocm-dev is not part of rocm usecase, 
2.fix ubuntu names
3.fix ubuntu version side issues to remove build numbers
4. PACKAGELIST being  a pre processor directive does not get correct values, make the definition part of code


rcqt.so/src/action.cpp
```diff
@@ -85,8 +84,8 @@ using std::vector;
 
 // Set Operating System variant
 const OSType rcqt_action::os_type = getOS();
-
 rcqt_action::rcqt_action() {
+  PACKAGELIST = getOS() == OSType::Ubuntu?"debpackagelist":"rpmpackagelist";
```
jkottiku, line 88: Just put round brackets for readability.

jkottiku (approved):
Looks good.

PR #662, opened 2023-10-18
Power API changes et al.

jkottiku:
1. Change smi get power API in IET & GM.
2. Add GM & SMQT configuration.
3. Bug fixes in GM module.
4. Junction temp. instead of edge temp.
5. Whole number display of power in Watts.
6. Correct fan% calculation in GM.

### DIFF
--- a/iet.so/src/iet_worker.cpp
+++ b/iet.so/src/iet_worker.cpp
@@ -209,7 +209,7 @@ bool IETWorker::do_iet_power_stress(void) {
     std::chrono::time_point<std::chrono::system_clock> iet_start_time, end_time,
         sampling_start_time;
     uint64_t  total_time_ms;
-    uint64_t  last_avg_power;
+    uint64_t  last_power;
     string    msg;
     float     cur_power_value = 0;
     float     totalpower = 0;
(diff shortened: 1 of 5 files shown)


iet.so/src/iet_worker.cpp
```diff
@@ -228,12 +229,13 @@ bool IETWorker::do_iet_power_stress(void) {
...
         if (rvs::lp::Stopping())
             break;
-        // get GPU's current average power
-        rsmi_status_t rmsi_stat = rsmi_dev_power_ave_get(smi_device_index , 0,
-                &last_avg_power);
 
+        cur_power_value = 0;
+
+        // get GPU's current average power
+        rsmi_status_t rmsi_stat = rsmi_dev_power_get(smi_device_index, &last_power, &type);
```
manoj-freyr, line 236: do we need to check "type" value here?

rvs/conf/smqt_single.conf
```diff
@@ -0,0 +1,41 @@
+# ################################################################################
+# #
```
manoj-freyr, line 2: how are the values of variables decided?

rvs/conf/gm_single.conf
```diff
@@ -0,0 +1,39 @@
...
+# # FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL THE
+# # AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# # SOFTWARE.
+# #
+# ###############################################################################
+
+actions:
+- name: metrics_monitor
```
manoj-freyr, line 27: how are values of variables decided?

iet.so/src/iet_worker.cpp
jkottiku: Not really because not much use in RVS as **type** will differ in MI300 vs Older GPUs. 
MI300 - RSMI_CURRENT_POWER - Power based on device
MI200 & others - RSMI_AVERAGE_POWER - Power based on GPU die

https://confluence.amd.com/display/BLTZ/MI-300+ROCm-SMI+Features

rvs/conf/gm_single.conf
jkottiku: This is just a placeholder configuration file.
Some of bound values are just samples and will depend on the platform. We might have to create separate gm conf for each platform.

manoj-freyr (approved):
looks good

PR #664, opened 2023-10-23
Navi31 specific conf files

manoj-freyr:


jkottiku (approved):
See if you need to have gm conf. as well ? Some parameter in it are platform specific.
Other than that looks good.

PR #665, opened 2023-10-31
Changes in TST module

jkottiku:
1. Check junction temp. reaches target temp.
2. Changes in tst configuration.


manoj-freyr (approved):
looks good

PR #669, opened 2023-11-09
Handle invalid/empty conf file

jkottiku:
Exit RVS execution with error when input conf. file is either empty or invalid.


manoj-freyr (approved):
Looks good

PR #679, opened 2023-11-29
GPU hip index to smi index mapping

jkottiku:


src/gpu_util.cpp
```diff
@@ -269,19 +269,68 @@ void gpu_get_all_domain_id(std::vector<uint16_t>* pgpus_domain_id,
...
   return true;
 }
 
+/**
+ * @brief Get GPU smi index from hip index.
+ * @param hip_index GPU hip index
+ * @param smi_index GPU smi index
+ * @return 0 if successful, -1 otherwise
+ **/
+int gpu_hip_to_smi_index(int hip_index, uint32_t* smi_index) {
```
manoj-freyr, line 295: bool can be sensible return?

manoj-freyr, line 295: can we reuse, hip_to_smi_indices()'s map to do a quicker lookup instead of calculating with a fucntion call every time? we already have create da map, can move map to gpu_util or rvs_util and do a quick lookup instead of a function call over and over?

manoj-freyr (approved):
check if comments are simpler to implement in this iteration, if not take this as conditional approval as functionality is correct,

src/gpu_util.cpp
jkottiku: Just sticked to the standard followed by other functions in the file.

jkottiku: Yes, true. For now I wanted to keep the changes minimal and also keep gst as it is. Will design the API in more optimized manner afterwards.

PR #680, opened 2023-11-29
GPU hip index to smi index mapping

jkottiku:
Fix for erroneous power reporting in IET module

### DIFF
--- a/src/gpu_util.cpp
+++ b/src/gpu_util.cpp
@@ -34,6 +34,10 @@
 
 #include "rocm_smi/rocm_smi.h"
 #include "include/gpu_util.h"
+#define __HIP_PLATFORM_HCC__
+#include "hip/hip_runtime.h"
+#include "hip/hip_runtime_api.h"
+
 std::vector<uint16_t> rvs::gpulist::location_id;
 std::vector<uint16_t> rvs::gpulist::gpu_id;
 std::vector<uint16_t> rvs::gpulist::device_id;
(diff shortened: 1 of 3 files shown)


src/gpu_util.cpp
```diff
@@ -34,6 +34,10 @@
 
 #include "rocm_smi/rocm_smi.h"
 #include "include/gpu_util.h"
+#define __HIP_PLATFORM_HCC__
```
frepaul, line 37: This is deprecated we should ne using HIP_PLATFORM_AMD

frepaul (changes requested):

src/gpu_util.cpp
jkottiku: __HIP_PLATFORM_HCC__ is present in other code as well. Need to make this change separately.

src/gpu_util.cpp
```diff
@@ -269,19 +269,68 @@ void gpu_get_all_domain_id(std::vector<uint16_t>* pgpus_domain_id,
...
+ * @param hip_index GPU hip index
+ * @param smi_index GPU smi index
+ * @return 0 if successful, -1 otherwise
+ **/
+int gpu_hip_to_smi_index(int hip_index, uint32_t* smi_index) {
+
+  int hip_num_gpu_devices = 0;
+  uint32_t smi_num_devices = 0;
+  uint64_t val_ui64 = 0;
+  std::map<uint64_t, int> smi_map;
```
manoj-freyr, line 300: this can atleast be made a singleton/static to avoid repeat runs .. Approving for release block, note as todo

manoj-freyr (approved):
looks good,

PR #684, opened 2023-12-14
Using domain field in getting PCIe id

manoj-freyr:
BDF gives common number unless we use domain too to enumerate PCIe
    devices, which resulted in sending GEMM to one GPU and smi query to
    another. Hence using D:B:D value as uint 64, following smi conventions

### DIFF
--- a/iet.so/src/action.cpp
+++ b/iet.so/src/action.cpp
@@ -413,8 +413,9 @@ void iet_action::hip_to_smi_indices(void) {
 
         // compute device location_id (needed to match this device
         // with one of those found while querying the pci bus
-        uint16_t hip_dev_location_id =
-            ((((uint16_t) (props.pciBusID)) << 8) | (((uint16_t)(props.pciDeviceID)) << 3) );
+        uint64_t hip_dev_location_id = ( ( ((uint64_t)props.pciDomainID & 0xffff ) << 32) |
+            (((uint64_t) props.pciBusID & 0xff ) << 8) | (((uint64_t)props.pciDeviceID & 0x1f ) << 3) );
+
         if(smi_map.find(hip_dev_location_id) != smi_map.end()){
             hip_to_smi_idxs.insert({i, smi_map[hip_dev_location_id]});
         }


iet.so/src/action.cpp
```diff
@@ -413,8 +413,9 @@ void iet_action::hip_to_smi_indices(void) {
 
         // compute device location_id (needed to match this device
         // with one of those found while querying the pci bus
-        uint16_t hip_dev_location_id =
-            ((((uint16_t) (props.pciBusID)) << 8) | (((uint16_t)(props.pciDeviceID)) << 3) );
+        uint64_t hip_dev_location_id = ( ( ((uint64_t)props.pciDomainID & 0xffff ) << 32) |
```
jkottiku, line 416: Why don't add (OR) function bits as well ? for its completeness.

jkottiku (changes requested):

iet.so/src/action.cpp
jkottiku: Left shifting **_props.pciDomainID_** by 16 bits should be enough. Right ?

jkottiku (changes requested):

iet.so/src/action.cpp
manoj-freyr: Since we are calling the id to index match based on smi APIs, I am using the same pattern as SMI uses to generate a id from Domain:Bus:Device enumerations. Also, strictly speaking we need to have another part for Function, which always if 0 in our case and hence we ignored it all this while and only 16 bit number was sufficient to encode B:D values. Now we are considering domain too , using 64 bit integer makes sense and using top 32 bits for domain is more sensible and gives bottom 32 bits for B:D:F values(9+ 6+ 4 = 19 bits which when aligned becomes 32 bits).

manoj-freyr: I dont get the OR 3 function bits part? We are ORing 3 values alright here?

jkottiku: > I dont get the OR 3 function bits part? We are ORing 3 values alright here?

I meant function bits in Bus:Device.Function format. Seems hip interface is not providing it. So, we can ignore it for now.

jkottiku: > Since we are calling the id to index match based on smi APIs, I am using the same pattern as SMI uses to generate a id from Domain:Bus:Device enumerations. Also, strictly speaking we need to have another part for Function, which always if 0 in our case and hence we ignored it all this while and only 16 bit number was sufficient to encode B:D values. Now we are considering domain too , using 64 bit integer makes sense and using top 32 bits for domain is more sensible and gives bottom 32 bits for B:D:F values(9+ 6+ 4 = 19 bits which when aligned becomes 32 bits).

Ok. Let us follow the same format for PCIe id as smi.

jkottiku (approved):

iet.so/src/action.cpp
jkottiku: From rsmi_dev_pci_id_get() API documentation, it seems that for generating bdfid smi is using function bits as well. Need to check how are they able to get it as HPI API doesn't provide it.

manoj-freyr: one can get these values from sys fs entry for that device's properties

PR #687, opened 2024-01-08
1. Added PCI BDF in pebb & pbqt logs

jkottiku:
2. Indentation & other minor log updates

### DIFF
--- a/include/gpu_util.h
+++ b/include/gpu_util.h
@@ -42,7 +42,7 @@ extern void gpu_get_all_domain_id(std::vector<uint16_t>* pgpus_domain_id,
                 std::map<std::pair<uint16_t, uint16_t> , uint16_t>& pgpus_dom_loc_map); 
 extern bool gpu_check_if_mcm_die (int idx);
 extern int gpu_hip_to_smi_index(int hip_index, uint32_t* smi_index);
-extern void gpu_get_all_pci_bdf(std::vector<std::string>* ppci_bdf);
+extern void gpu_get_all_pci_bdf(std::vector<std::string>& ppci_bdf);
 
 namespace rvs {
 
--- a/include/gpu_util.h
+++ b/include/gpu_util.h
@@ -42,10 +42,9 @@ extern void gpu_get_all_domain_id(std::vector<uint16_t>* pgpus_domain_id,
                 std::map<std::pair<uint16_t, uint16_t> , uint16_t>& pgpus_dom_loc_map); 
 extern bool gpu_check_if_mcm_die (int idx);
 extern int gpu_hip_to_smi_index(int hip_index, uint32_t* smi_index);
+extern void gpu_get_all_pci_bdf(std::vector<std::string>* ppci_bdf);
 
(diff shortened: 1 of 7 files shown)


pbqt.so/src/action.cpp
```diff
@@ -480,18 +480,36 @@ int pbqt_action::create_threads() {
         return -1;
       }
 
+      std::string srcpcibdf;
```
manoj-freyr, line 483: keeping with other larger variable names, see if you can use snake case for these variable names?

include/gpu_util.h
```diff
@@ -42,11 +42,10 @@ extern void gpu_get_all_domain_id(std::vector<uint16_t>* pgpus_domain_id,
                 std::map<std::pair<uint16_t, uint16_t> , uint16_t>& pgpus_dom_loc_map); 
 extern bool gpu_check_if_mcm_die (int idx);
 extern int gpu_hip_to_smi_index(int hip_index, uint32_t* smi_index);
+extern void gpu_get_all_pci_bdf(std::vector<std::string>* ppci_bdf);
```
manoj-freyr, line 45: please use reference instead of pointers , will make it more null safe and general good practice. atleast new APIs can have this way instead of following old ones I feel

src/gpu_util.cpp
```diff
@@ -264,6 +266,84 @@ void gpu_get_all_domain_id(std::vector<uint16_t>* pgpus_domain_id,
...
+
+          continue;
+        }
+      }
+
+      /* Form PCI BDF */
+      (*ppci_bdf).push_back(domain + ":" + bus + ":" + device + ".0" );
+    }
+
+    f_id.close();
```
manoj-freyr, line 342: destructor closes it, so do we need explicit close?

pbqt.so/src/action.cpp
jkottiku: True, followed the convention already in place. Better we stick with it for now.

include/gpu_util.h
jkottiku: Yes, will use reference. GPU APIs have used pointers and references already.

src/gpu_util.cpp
jkottiku: Do we have destructor here ? As the methods all are static right ?
Also, closing stream inside loop looks clearer and followed as convention in other similar APIs as well.

manoj-freyr: Whole RAII ensures cleanup happens before we take object out of scope. methods are static but f_id is local variable in this scope and when it encounters braces its destructor is called for sure, which closes the stream object.

jkottiku: In that case, f_id & f_prop have to be defined inside the for loop. right ?

manoj-freyr (approved):

src/gpu_util.cpp
jkottiku: As discussed, for now will keep as it is.

PR #702, opened 2024-02-08
[GST] Support for data types based GEMM operations

jkottiku:


gst.so/src/gst_worker.cpp
```diff
@@ -250,20 +241,11 @@ bool GSTWorker::do_gst_ramp(int *error, string *err_description) {
...
-        std::unique_lock<std::mutex> lk(mutex);
-        cv.wait(lk);
-
-        if(!blas_status) {
-          msg = "[" + action_name + "] " + MODULE_NAME + " " +
-            std::to_string(gpu_id) + " " + " BLAS gemm operations failed !!! ";
-          rvs::lp::Log(msg, rvs::logtrace);
-        }
+        // Wait for GEMM operation to complete
+        gpu_blas->is_gemm_op_complete();
```
manoj-freyr, line 248: this is immediate returning call, it might return false/true based on hipStreamQuery(hip_stream)  complete or not. we are not checking return here and incomplete gemm operation will still return above call. check in a loop for proper working

gst.so/src/gst_worker.cpp
```diff
@@ -456,49 +440,66 @@ bool GSTWorker::do_gst_stress_test(int *error, std::string *err_description) {
...
+            rvs::lp::Log(msg, rvs::logtrace);
+          }
         }
+        else {
+          // data type (fp8 & bp16) based gemm operations
 
-        //End the timer
-        end_time = gpu_blas->get_time_us();
+          // Wait for GEMM operation to complete
+          gpu_blas->is_gemm_op_complete();
```
manoj-freyr, line 468: same as aboce, check in a loop to ensure completion as this is a non blocking call

gst.so/src/gst_worker.cpp
```diff
@@ -542,9 +543,19 @@ void GSTWorker::run() {
             " Starting the GST stress test "; 
     rvs::lp::Log(msg, rvs::logtrace);
 
+    // log GST ramp up - start message
+    msg = "[" + action_name + "] " + MODULE_NAME + " " +
+            std::to_string(gpu_id) + " Start of GPU ramp up";
+    rvs::lp::Log(msg, rvs::logresults);
+
     // let the GPU ramp-up and check the result
     bool ramp_up_success = do_gst_ramp(&error, &err_description);
```
manoj-freyr, line 552: can we use the return value too to determine success/failure?

tst.so/src/tst_worker.cpp
```diff
@@ -154,7 +154,7 @@ void TSTWorker::blasThread(int gpuIdx, uint64_t matrix_size, std::string tst_ops
     gem_ops = 0;
    // setup rvsBlas
     gpu_blas = std::unique_ptr<rvs_blas>(new rvs_blas(gpuIdx,  matrix_size,  matrix_size,  matrix_size, transa, transb, alpha, beta, 
-          tst_lda_offset, tst_ldb_offset, tst_ldc_offset, tst_ops_type));
+          tst_lda_offset, tst_ldb_offset, tst_ldc_offset, tst_ops_type, ""));
```
manoj-freyr, line 157: instead of this, we can have default param for type as empty string in prototype of rvs_blas?

src/rvs_blas.cpp
```diff
@@ -480,22 +610,70 @@ bool rvs_blas::run_blass_gemm(std::string ops_type) {
...
+      if (rocblas_gemm_ex(blas_handle, transa, transb,
+            rvs_blas::m, rvs_blas::n, rvs_blas::k, &alpha,
+            dda, a_type, blas_lda_offset,
+            ddb, b_type, blas_ldb_offset, &beta,
+            ddc, c_type, blas_ldc_offset,
+            ddd, d_type, blas_ldc_offset,
+            compute_type, algo, sol_index, flags) != rocblas_status_success) {
+
+        is_error = true;  // GPU cannot enqueue the gemm
+        std::cout << "\n Error in rocblas_gemm_ex() !!!" << "\n";
```
manoj-freyr, line 671: i see you are following prints as in file, cout is race-prone compared to rvs::log , if not in this draft, have a todo to replace?

manoj-freyr (changes requested):
check the comments please

gst.so/include/action.h
```diff
@@ -61,6 +61,7 @@ class gst_action: public rvs::actionbase {
     virtual int run(void);
     static void cleanup_logs();
     std::string gst_ops_type;
+    std::string gst_data_type;
```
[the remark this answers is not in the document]
jkottiku: Good catch. Added gst_data_type alongside gst_ops_type initially. Moved both to private now.

gst.so/src/gst_worker.cpp
jkottiku: is_gemm_op_complete() call is blocking now as we use hipStreamSynchronize() instead of hipStreamQuery().
Will add return checking for both run_blass_gemm() and is_gemm_op_complete().

gst.so/src/gst_worker.cpp
jkottiku: is_gemm_op_complete() call is blocking now as we use hipStreamSynchronize() instead of hipStreamQuery(). So, no busy loop required.

src/rvs_blas.cpp
jkottiku: Yes, we have to move to rvs::log completely. There other modules as well.

gst.so/src/gst_worker.cpp
jkottiku: Currently, we have not tuned ramp up time (using default time now) yet. Better not to provide ramp-up as a result which might cause ambiguity even though most of the time it will be SUCCESS.

tst.so/src/tst_worker.cpp
jkottiku: We will have to still pass complete set of arguments to the rvs_blas constructor Right ?

manoj-freyr (approved):
looks good

PR #705, opened 2024-02-09
Provide an option to fetch ROCm install path using the api provided by rocm-core

raramakr:
By default use the existing way of getting ROCM_PATH from build arguments.  This will help in backward compatibility. 
Build scripts can enable the compile time option, so that api provided by rocm-core can be used to fetch ROCm Install path.


jkottiku (approved):
Build looks good.

PR #707, opened 2024-02-14
Use rocblas_float instead of _Float32

jkottiku:


manoj-freyr (approved):
looks good, since roc* API must work on all OS

PR #710, opened 2024-02-16
SWDEV-446072 - Correct the package dependency list usage (#708)

raramakr:
rocm-core was added to the package dependency list on a conditional basis. While disabling the dependency, the package Depends list was getting populated incorrectly and resulting in installation failure. Corrected the same


nunnikri:
Approved by release meeting held on 2/22/23

PR #715, opened 2024-02-28
MI300X custom pbqt conf.

jkottiku:
pebb & pbqt result log indentation.

### DIFF
--- a/pbqt.so/src/action.cpp
+++ b/pbqt.so/src/action.cpp
@@ -408,14 +408,18 @@ bool pbqt_action::get_all_common_config_keys(void) {
  *
  * */
 int pbqt_action::create_threads() {
-  std::string msg;
 
+  std::string msg;
   std::vector<uint16_t> gpu_id;
   std::vector<uint16_t> gpu_device_id;
   uint16_t transfer_ix = 0;
   bool bmatch_found = false;
+  char srcgpuid_buff[12];
+  char dstgpuid_buff[12];
+
   gpu_get_all_gpu_id(&gpu_id);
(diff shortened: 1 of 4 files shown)


pbqt.so/src/action.cpp
```diff
@@ -508,9 +512,13 @@ int pbqt_action::create_threads() {
       // perform peer check
       if (is_peer(gpu_id[i], gpu_id[j])) {
         RVSTRACE_
+
+        snprintf(srcgpuid_buff, sizeof(srcgpuid_buff), "%5d", gpu_id[i]);
```
manoj-freyr, line 516: see if you can use std::setw() instead of %5d ?

pbqt.so/src/action.cpp
```diff
@@ -881,10 +897,14 @@ int pbqt_action::print_final_average() {
     transfer_ix = (*it)->get_transfer_ix();
     transfer_num = (*it)->get_transfer_num();
 
+    snprintf(transfer_buff, sizeof(transfer_buff), "%2d", transfer_ix);
```
manoj-freyr, line 900: can we use setw() when printing instead of %d qualifiers?

manoj-freyr (changes requested):

pbqt.so/src/action.cpp
jkottiku: Yes, I tried it. But we need addition steps made the code kind of complex. This implementation easily fit into the code (especially **_msg_** log string) we have in place.

manoj-freyr (approved):
looks good

manoj-freyr (approved):

PR #721, opened 2024-03-14
1. Set gemm matrix size correctly.

jkottiku:
2. Remove sleep at intervals during test duration.

### DIFF
--- a/iet.so/src/iet_worker.cpp
+++ b/iet.so/src/iet_worker.cpp
@@ -130,7 +130,7 @@ void IETWorker::log_interval_gflops(double gflops_interval) {
 
 void IETWorker::blasThread(int gpuIdx,  uint64_t matrix_size, std::string  iet_ops_type, 
     bool start, uint64_t run_duration_ms, int transa, int transb, float alpha, float beta,
-    int iet_lda_offset, int iet_ldb_offset, int iet_ldc_offset){
+    int iet_lda_offset, int iet_ldb_offset, int iet_ldc_offset) {
 
     std::chrono::time_point<std::chrono::system_clock> iet_start_time, iet_end_time;
     double timetakenforoneiteration;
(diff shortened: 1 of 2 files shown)


iet.so/src/iet_worker.cpp
```diff
@@ -219,7 +212,7 @@ bool IETWorker::do_iet_power_stress(void) {
     rvs::action_result_t action_result;
     RSMI_POWER_TYPE type = RSMI_INVALID_POWER;
 
-    std::thread t(&IETWorker::blasThread, this, gpu_device_index, matrix_size_a, iet_ops_type, start, run_duration_ms,
+    std::thread t(&IETWorker::blasThread, this, gpu_device_index, matrix_size, iet_ops_type, start, run_duration_ms,
```
manoj-freyr, line 215: good catch

iet.so/src/iet_worker.cpp
```diff
@@ -176,26 +177,18 @@ void IETWorker::blasThread(int gpuIdx,  uint64_t matrix_size, std::string  iet_o
...
         //Print the gflops interval
         log_interval_gflops(gflops_interval);
+
         // check end test to avoid unnecessary sleep
         if (endtest)
             break;
-        //if gemm ops greater than 10000, lets yield
-        //if this is not happening we are ending up in
-        //out of memmory state
-        if(gem_ops > 10000) {
```
manoj-freyr, line 193: why are we removing ?

iet.so/src/iet_worker.cpp
```diff
@@ -176,26 +177,18 @@ void IETWorker::blasThread(int gpuIdx,  uint64_t matrix_size, std::string  iet_o
         iet_end_time = std::chrono::system_clock::now();
         //Duration in the call
         duration = time_diff(iet_end_time, iet_start_time);
-        gem_ops++;
```
manoj-freyr, line 179: why are we removing?

iet.so/src/iet_worker.cpp
manoj-freyr: but what happens if we specify martix sizes a b and c, how can we ignore it ?

iet.so/src/iet_worker.cpp
jkottiku: This sleep results in unsustained GPU power utilization resulting in fluctuating power usage which is not expected why testing max. power cap testing.

iet.so/src/iet_worker.cpp
manoj-freyr: then this is a todo as a,b,and c can be used to override behavior and there is no need to have a b and c to be of same dimension

manoj-freyr (approved):

PR #730, opened 2024-03-23
Port babel to AMD internal HIPStream version

jkottiku:
Back port babel to AMD internal Babelstream (HIPStream) version of 3.4


manoj-freyr (approved):
looks good.

PR #733, opened 2024-03-28
Support for matrix data init. types

jkottiku:
- Support for new random integer and trignometric float values added in rvs_blas.
- New parameter matrix init. type added in gst conf. file with trig and rand values.
- Updated MI300X gst_single.conf with both trignometric float and random interger tests for fp8, fp16, bf16, sgemm and dgemm.
- Removed decimal in GFLOPS from the output logs.
- Other minor logging changes.

### DIFF
--- a/gst.so/include/action.h
+++ b/gst.so/include/action.h
@@ -87,6 +87,9 @@ class gst_action: public rvs::actionbase {
     uint64_t gst_matrix_size_b;
     uint64_t gst_matrix_size_c;
 
+    //! matrix initialization
+    std::string gst_matrix_init;
+
     //Parameter to heat up
     uint64_t gst_hot_calls;
 
--- a/gst.so/include/action.h
+++ b/gst.so/include/action.h
@@ -69,8 +69,8 @@ class gst_action: public rvs::actionbase {
     uint64_t gst_ramp_interval;
     //! maximum allowed number of target_stress violations
     int gst_max_violations;
-    //! specifies whether to copy the matrices to the GPU before each
-    //! SGEMM operation
+    //! specifies whether to copy the matrices
+    //! to the GPU before each GEMM operation
     bool gst_copy_matrix;
     //! target stress (in GFlops) that the GPU will try to achieve
     float gst_target_stress;
@@ -78,16 +78,17 @@ class gst_action: public rvs::actionbase {
     //! the ramp period for the test to succeed)
     float gst_tolerance;
 
-    //Alpha and beta value
-    float      gst_alpha_val;
-    float      gst_beta_val;
+    //! Alpha and beta value
+    float gst_alpha_val;
+    float gst_beta_val;
 
-    //! matrix size for SGEMM
+    //! matrix sizes for GEMM operation
     uint64_t gst_matrix_size_a;
     uint64_t gst_matrix_size_b;
     uint64_t gst_matrix_size_c;
 
-    //! matrix initialization
+    //! matrix initialization method :
+    //! default, random integer or trignometric float
     std::string gst_matrix_init;
 
     //Parameter to heat up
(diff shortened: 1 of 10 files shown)


gst.so/include/action.h
```diff
@@ -87,6 +87,9 @@ class gst_action: public rvs::actionbase {
     uint64_t gst_matrix_size_b;
     uint64_t gst_matrix_size_c;
 
+    //! matrix initialization
```
manoj-freyr, line 90: state in comments what it actually does

rvs/conf/MI300X/gst_single.conf
```diff
@@ -24,94 +24,196 @@
...
   ramp_interval: 5000
-  duration: 30000
+  duration: 15000
   copy_matrix: false
-  target_stress: 1000000
+  target_stress: 900000
   matrix_size_a: 8192
   matrix_size_b: 8192
   matrix_size_c: 8192
+  matrix_init: trig
```
manoj-freyr, line 38: mention the usage detail at top of the conf file as a comment

src/rvs_blas.cpp
```diff
@@ -26,11 +26,48 @@
...
+#include <cmath>
+#include <random>
+#include <thread>
+
+/* ============================================================================================ */
+// Random number generator
+using rvsblas_rng_t = std::mt19937;
+
+// Random number generator
+rvsblas_rng_t rvsblas_seed(69069); // A fixed seed to start at
```
manoj-freyr, line 38: do we need not randomise this initial seed?

src/rvs_blas.cpp
```diff
@@ -758,91 +797,129 @@ void rvs_blas::generate_random_matrix_data(void) {
...
  * @brief fast pseudo random generator 
  * @return floating point random number
  */
-float rvs_blas::fast_pseudo_rand(uint64_t *nextr) {
+float rvs_blas::fast_pseudo_rand(uint64_t *nextr, size_t i) {
+
+  if ("rand" == matrix_init) {
+
+    if("fp8_r" == data_type) {
+      return (float)std::uniform_int_distribution<int>(1, 2)(rvsblas_t_rng);
```
manoj-freyr, line 901: why parametrizing as int? wont it be only integral values1 and 2 which we are converting to float? will it make more sense to parametrizing it as float?

src/rvs_blas.cpp
```diff
@@ -758,91 +797,129 @@ void rvs_blas::generate_random_matrix_data(void) {
...
+    else if("bf16_r" == data_type)
+    {
+      return (float)std::uniform_int_distribution<int>(-2, 2)(rvsblas_t_rng);
+    }
+    else { /* sgemm, dgemm */
+      return rvsblas_uniform_int_1_10();
+    }
+  }
+  else if ("trig" == matrix_init) {
+    return static_cast<float>(sin(i));
```
manoj-freyr, line 916: better to cast argument to float than to cast return type? looks like arg, if int, will be treated as double which returns double as result and we are downcasting it to float in this case?

manoj-freyr (commented):
check the comments

gst.so/include/action.h
jkottiku: done

src/rvs_blas.cpp
jkottiku: This is just the main thread seed at init, other seeds are based on thread ids.

src/rvs_blas.cpp
jkottiku: Followed the same logic as in rocblas-bench.

src/rvs_blas.cpp
jkottiku: Done

rvs/conf/MI300X/gst_single.conf
jkottiku: Done

manoj-freyr (approved):
looks good, attach the results in the ticket to see the difference in perf numbers

jkottiku:
Test logs attached in the ticket.

PR #739, opened 2024-04-22
updates

manoj-freyr:
Fixes for versioning


rcqt.so/src/debPackageInfo.cpp
```diff
@@ -1,24 +1,68 @@
...
+        if (hyp != std::string::npos){
+          pvers = pvers.substr(0,hyp);
+        }
+	if (pat.compare("(>=") == 0)
+          pvers+="+";
+    }
+  } 
+  return {remSpaces(pname), remSpaces(pvers)};
+}
+/*
```
jkottiku, line 44: Lets remove the old implementation ?

rcqt.so/src/debPackageInfo.cpp
```diff
@@ -1,24 +1,68 @@
 #include <iostream>
+#include <vector>
 #include <sstream>
 #include <fstream>
 #include "include/debPackageInfo.h"
 
+std::vector<std::string> dependency_patterns{"(=", "(>="};
+
+std::string findPattern(std::string word){
```
jkottiku, line 9: Should we move this function to utils ? Or is this really specific to deb ?

jkottiku (changes requested):

rcqt.so/src/debPackageInfo.cpp
manoj-freyr: did not see any such versions on rhel or sles. so currently kept in ubuntu/deb

jkottiku (approved):

PR #740, opened 2024-04-22
use function in identifying end point

manoj-freyr:


jkottiku:
Just check if we have missed function elsewhere as well ...

jkottiku (approved):

src/gpu_util.cpp
```diff
@@ -627,3 +627,58 @@ int rvs::gpulist::gpu2domain(const uint16_t GpuID, uint16_t* pDomain) {
   return 0;
 }
 
+/* common function*/
+bool rvs::gpulist::availableGPUs(std::map<int, uint16_t> &gpus_device_index, bool property_all,
```
jkottiku, line 631: Are we using this function internally ?

manoj-freyr: remvoed

jkottiku (approved):

PR #741, opened 2024-04-23
Power sampling interval corrections

jkottiku:
update logging power

### DIFF
--- a/iet.so/src/iet_worker.cpp
+++ b/iet.so/src/iet_worker.cpp
@@ -225,7 +225,7 @@ bool IETWorker::do_iet_power_stress(void) {
 
         cur_power_value = 0;
 
-        // get GPU's current average power
+        // get GPU's current/average power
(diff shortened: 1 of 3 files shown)


iet.so/src/iet_worker.cpp
```diff
@@ -258,7 +258,7 @@ bool IETWorker::do_iet_power_stress(void) {
         }
 
         //It doesnt make sense to read power continously so slowing down
-        sleep(1000);
+        sleep(sample_interval);
```
manoj-freyr, line 261: Would we need to use this_thread::sleep() instead now that we are making a fine grained change? sleep() is not thread aware.

manoj-freyr (changes requested):

iet.so/src/iet_worker.cpp
jkottiku: sleep_for() used instead of usleep

manoj-freyr (approved):
looks good

PR #744, opened 2024-05-03
gpup run need not stop at individual error

manoj-freyr:
for some GPUs if said properties are absent, we should report it and move on to next property instead of prematurely abandoning the rest of the flow.


gpup.so/src/action.cpp
```diff
@@ -478,7 +481,7 @@ int gpup_action::run(void) {
 
       if (sts) {
```
jkottiku, line 482: Move this checking after **_sts = property_get_value(*it);_** line no. 467

jkottiku (changes requested):

jkottiku (approved):

PR #760, opened 2024-06-18
common config. read cleanup

manoj-freyr:
common configs are read and parsed in each module thought most of it is redundant.
This PR organises common config parsing in actionbase class and only overrides where a specific subset is needed.
For future modules, unless we need specific params, we must reuse this function and add others in module specific functions


jkottiku (approved):
Approved. Make sure all the modules are tested.

PR #766, opened 2024-06-26
Gemm self & accuracy check feature

jkottiku:
1. Validate gemm output for consistency/repeatability by self-check (current vs previous output).
2. Validate gemm output for accuracy by accuracy-check (device (GPU) output vs host (CPU) output).

### DIFF
--- a/gst.so/src/gst_worker.cpp
+++ b/gst.so/src/gst_worker.cpp
@@ -45,7 +45,6 @@
 #define GST_BYTES_COPIED_PER_OP_OUTPUT_KEY      "bytes_copied_per_op"
 #define GST_TRY_OPS_PER_SEC_OUTPUT_KEY          "try_ops_per_sec"
 
-#define GST_LOG_GFLOPS_INTERVAL_KEY             "GFLOPS"
 #define GST_LOG_SELF_CHECK_ERROR_KEY            "self-check error"
 #define GST_LOG_ACCU_CHECK_ERROR_KEY            "accu-check error"
 #define GST_LOG_GFLOPS_INTERVAL_KEY             "GFLOPS"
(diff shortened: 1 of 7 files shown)


gst.so/src/gst_worker.cpp
```diff
@@ -489,6 +498,38 @@ bool GSTWorker::do_gst_stress_test(int *error, std::string *err_description) {
...
+    if (self_check || accu_check) {
+
+      if (error_inject) {
+        gpu_blas->set_gemm_error(error_freq, error_count);
+      }
+
+      double self_error = 0.0;
+      double accu_error = 0.0;
+
+      gpu_blas->validate_gemm(self_check, accu_check, self_error, accu_error);
```
manoj-freyr, line 510: unlikely case , but we must ensure atleast 2 runs of this call to compare current and previous results.  how to ensure this ?

include/rvs_blas.h
```diff
@@ -219,6 +237,12 @@ class rvs_blas {
     void release_host_matrix_mem(void);
     float fast_pseudo_rand(uint64_t *nextr, size_t i);
 
+    template <typename T>
+      bool check_result_consistency(void * dout, uint64_t size, double &error);
```
manoj-freyr, line 241: size_t is always a better type for this than uint

gst.so/src/gst_worker.cpp
jkottiku: In src/rvs_blas.cpp, function check_result_consistency() check for comment /* Exit first iteration of self-check as there is no previous result yet ! */ its done here.

manoj-freyr (approved):
Update decsriptions

PR #770, opened 2024-07-02
1. Added data type, matrix init and hot calls.

jkottiku:
2. Set matrix sizes properly in IET module.
3. iet workload thread functions definition changes.
4. rvs_blas interfaces function definition changes.
5. Remove redundant code.

### DIFF
--- a/iet.so/src/iet_worker.cpp
+++ b/iet.so/src/iet_worker.cpp
@@ -114,73 +114,66 @@ IETWorker::IETWorker():endtest(false) {
 IETWorker::~IETWorker() {
 }
 
-
-
-/**
- * @brief logs the Gflops computed over the last log_interval period
- * @param gflops_interval the Gflops that the GPU achieved
- */
-void IETWorker::log_interval_gflops(double gflops_interval) {
-    string msg;
-    msg = " GPU flops :" + std::to_string(gflops_interval);
-    rvs::lp::Log(msg, rvs::logtrace);
-    log_to_json(IET_LOG_GFLOPS_INTERVAL_KEY, std::to_string(gflops_interval),
-                rvs::loginfo);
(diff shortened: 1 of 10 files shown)


iet.so/src/iet_worker.cpp
```diff
@@ -114,73 +114,66 @@ IETWorker::IETWorker():endtest(false) {
...
     std::chrono::time_point<std::chrono::system_clock> iet_start_time, iet_end_time;
-    double timetakenforoneiteration;
-    double gflops_interval;
-    double duration;
-    uint64_t gem_ops;
+    double duration = 0;
     std::unique_ptr<rvs_blas> gpu_blas;
-    rvs_blas *free_gpublas;
-    string msg;
+    uint64_t _matrix_size_a;
```
manoj-freyr, line 122: inconsistent naming convention for local vars, see if you can have without starting underscores

iet.so/src/iet_worker.cpp
```diff
@@ -114,73 +114,66 @@ IETWorker::IETWorker():endtest(false) {
...
-        //Print the gflops interval
-        log_interval_gflops(gflops_interval);
-
-        // check end test to avoid unnecessary sleep
-        if (endtest)
-            break;
+        // run GEMM operation
+        if(!gpu_blas->run_blas_gemm()) {
+          endtest = true;
+          break;
```
manoj-freyr, line 159: this breaks only from inner loop, is this what you want, if blas fails ? or do we sensibly break out of while loop

jkottiku: Since we are setting endtest=true, it will exit the outer loop too.

manoj-freyr (approved):
looks good

PR #773, opened 2024-07-05
Strided batched gemm support

jkottiku:


gst.so/src/gst_worker.cpp
```diff
@@ -150,7 +151,7 @@ void GSTWorker::hit_max_gflops(int *error, string *err_description) {
 
     // Waits for GEMM operation to complete
     if(!gpu_blas->is_gemm_op_complete())
-      continue;
+      continue;  // failed to run the GEMM operation
```
manoj-freyr, line 154: dont we need to break out?

jkottiku: This function is obsolete. Hence, didn't change it.

manoj-freyr (approved):

PR #779, opened 2024-07-23
1. strided batched matrix init. (using hiprand)

jkottiku:
2. MI308X iet single conf.
3. MI308X iet thermal conf.

### DIFF
--- a/src/rvs_blas.cpp
+++ b/src/rvs_blas.cpp
@@ -104,6 +104,7 @@ rvs_blas::rvs_blas(int _gpu_device_index, int _m, int _n, int _k, std::string _m
   , hpo(nullptr), hco(nullptr)
   , hout(nullptr), hdout(nullptr)
   , hip_stream(nullptr)
+  , hiprand_generator(nullptr)
   , blas_handle(nullptr)
   , is_handle_init(false)
   , is_error(false)
@@ -540,7 +541,8 @@ void rvs_blas::release_gpu_matrix_mem(void) {
 
   if (is_handle_init) {
     rocblas_destroy_handle(blas_handle);
-    hiprandDestroyGenerator(hiprand_generator);
+    if(hiprand_generator)
+      hiprandDestroyGenerator(hiprand_generator);
     hipStreamDestroy(hip_stream);
   }
 }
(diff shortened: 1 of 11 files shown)


src/rvs_blas.cpp
```diff
@@ -218,6 +218,20 @@ bool rvs_blas::init_gpu_device(void) {
     return false;
   }
 
+  if("hiprand" == matrix_init) {
+
+    // Create hipRAND generator, assign stream.
+    if(hiprandCreateGenerator(&hiprand_generator, HIPRAND_RNG_PSEUDO_DEFAULT) != HIPRAND_STATUS_SUCCESS) {
+      std::cout << "\n hiprandCreateGenerator() failed !!!" << "\n";
+      return false;
```
manoj-freyr, line 226: all returns without cleaning up explicitly allocated values like stream or blas allocate will result in resource leakages. Maybe add as a TODO but error returns are not used to do any cleanup even in caller heirarchy

src/rvs_blas.cpp
```diff
@@ -916,86 +942,115 @@ bool rvs_blas::run_blas_gemm(void) {
...
 
-      for (i = 0; i < size_a; ++i)
-        ha[i] = fast_pseudo_rand(&nextr, i);
+      if(ops_type == "dgemm") {
 
-      for (i = 0; i < size_b; ++i)
-        hb[i] = fast_pseudo_rand(&nextr, i);
+        if(hiprandGenerateUniformDouble(hiprand_generator, ddbla, size_a) != HIPRAND_STATUS_SUCCESS) {
+          std::cout << "\n hiprandGenerateUniformDouble() failed !!!" << "\n";
+          return;
```
manoj-freyr, line 953: in case of any of these failures again return value not checked or used in caller and it proceeds for operations as if no error.(in iet and gst)

jkottiku: error set is function calls fails.

src/rvs_blas.cpp
jkottiku: hiprandDestroyGenerator() called in rvs_blas::release_gpu_matrix_mem() to release rand generator alongside rocblas handle and hip stream.

manoj-freyr (approved):
looks good

PR #784, opened 2024-07-26
enable checks to prevent infinite loop when pwr budget search

manoj-freyr:


src/pci_caps.cpp
```diff
@@ -424,7 +432,7 @@ void get_pwr_budgeting(struct pci_dev *dev, uint8_t pb_pm_state,
             }
 
             i++;
-        } while (1);
+        } while (i < BYTE_MAX); // DSR size is 1 byte, no need to run forever
```
jkottiku, line 435: Max. number of read is 255. right ?

jkottiku (approved):

src/pci_caps.cpp
```diff
@@ -40,6 +40,7 @@ extern "C" {
 #define PCI_CAP_DATA_MAX_BUF_SIZE 1024
 #define PCI_CAP_NOT_SUPPORTED "NOT SUPPORTED"
 #define MEM_BAR_MAX_INDEX 5
+#define BYTE_MAX 255
```
jkottiku, line 43: If 256 iterations are required, then this should be 256 right ?

PR #787, opened 2024-07-30
Add b2b, warm_calls & hot_calls parameters

jkottiku:


pbqt.so/src/worker.cpp
```diff
@@ -165,7 +165,7 @@ int pbqtworker::do_transfer() {
   for (size_t i = 0; brun && i < block_size.size(); i++) {
     current_size = block_size[i];
     sts = pHsa->SendTraffic(src_node, dst_node, current_size,
-                            bidirect, &duration);
+                            bidirect, false, 1, 1, &duration);
```
manoj-freyr, line 168: have local vars by some legible names related to b2b , warm and hot calls and then use it in the function call, else we need to look up the declaration of function to see what is false and 1

pebb.so/include/action.h
```diff
@@ -93,6 +93,14 @@ class pebb_action : public rvs::actionbase {
   //! link type
   int link_type;
   std::string link_type_string;
+
+  //! Number of warm calls (transfer iterations) before bandwidth calculation
```
manoj-freyr, line 97: a bit more detail on what is warm call, the reason for warm and hot calls, comments can state intent clearly

pebb.so/src/action.cpp
```diff
@@ -73,51 +73,71 @@ pebb_action::~pebb_action() {
...
+  }
+
+  error = property_get_int<uint32_t>(RVS_CONF_HOT_CALLS_KEY, &hot_calls, DEFAULT_HOT_CALLS);
+  if (error == 1) {
+    msg = "invalid '" + std::string(RVS_CONF_HOT_CALLS_KEY) + "' key";
+    rvs::lp::Err(msg, MODULE_NAME_CAPS, action_name);
+    bsts = false;
+  }
+
+  error = property_get_int<uint32_t>(RVS_CONF_WARM_CALLS_KEY, &warm_calls, DEFAULT_WARM_CALLS);
```
manoj-freyr, line 128: should there be a check on conditions for hot and warm calls? if one should be greater than the other etc?

jkottiku: warm and hot calls shall not be 0 - always default value.

manoj-freyr (approved):

PR #796, opened 2024-08-16
Updated iet conf. with multi power levels

jkottiku:


manoj-freyr (approved):

jkottiku:
This change is no more required.

PR #801, opened 2024-08-22
logging lib usage in babel instead of cout

manoj-freyr:
instead of cout, use thread safe log library

### DIFF
--- a/babel.so/src/rvs_stress.cpp
+++ b/babel.so/src/rvs_stress.cpp
@@ -14,6 +14,7 @@
 #include <iomanip>
 #include <cstring>
 #include <mutex>
+#include <sstream>
 
 #define VERSION_STRING "3.4"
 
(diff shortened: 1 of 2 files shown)


jkottiku (approved):

babel.so/src/rvs_stress.cpp
```diff
@@ -70,35 +71,37 @@ void run_stress(std::pair<int, uint16_t> device, int num_times, int ARRAY_SIZE,
 {
   std::string   msg;
   std::streamsize ss = std::cout.precision();
-
+  std::stringstream sstr;
   if (!output_as_csv)
   {
-    std::cout << "Running kernels " << num_times << " times" << std::endl;
+    msg = "Running kernels " + std::to_string(num_times) + " times, " ;
```
jkottiku, line 77: There is a "," at the end. Do we need it ?

babel.so/src/rvs_stress.cpp
```diff
@@ -70,35 +71,37 @@ void run_stress(std::pair<int, uint16_t> device, int num_times, int ARRAY_SIZE,
...
+    rvs::lp::Log(msg, rvs::logresults);
     if (mibibytes)
     {
       // MiB = 2^20
-      std::cout << std::setprecision(1) << std::fixed
+      sstr << std::setprecision(1) << std::fixed
                 << "Array size: " << ARRAY_SIZE*sizeof(T)*pow(2.0, -20.0) << " MiB"
-                << " (=" << ARRAY_SIZE*sizeof(T)*pow(2.0, -30.0) << " GiB)" << std::endl;
-      std::cout << "Total size: " << 3.0*ARRAY_SIZE*sizeof(T)*pow(2.0, -20.0) << " MiB"
+                << " (=" << ARRAY_SIZE*sizeof(T)*pow(2.0, -30.0) << " GiB), ";
```
jkottiku, line 91: There is a "," at the end. Do we need it ?

jkottiku (changes requested):

babel.so/src/rvs_stress.cpp
manoj-freyr: yeah to accomodate precision log which comes next, because using logging library we will be adding tag and timestamp for every new log line and i am trying to keep all relevent info in single log line for num of times and what precision

babel.so/src/rvs_stress.cpp
manoj-freyr: yes, it is a better delimiter than a newline since we will combine logs of one GPU as a virtually single log line

jkottiku (approved):

PR #828, opened 2024-10-21
Update doc conf version match regex

samjwu:
closes https://github.com/ROCm/ROCmValidationSuite/issues/827

fixes failing doc build due to mismatch


samjwu:
fixed in https://github.com/ROCm/ROCmValidationSuite/pull/823

PR #833, opened 2024-11-05
fix runtime symbol load issues for template

manoj-freyr:
Variadic Template function reference is not seen in shared object in nm or objdump but if i place it in specific actions it works.
WIll refactor this, pushing change to avoid run time so load failures.


jkottiku (approved):

jkottiku:
Let plan to move the template to the header later on instead of two same templates.

PR #862, opened 2024-12-11
configs for mi210

manoj-freyr:


rvs/conf/MI210/babel.conf
```diff
@@ -0,0 +1,51 @@
+# ################################################################################
+# #
+# # Copyright (c) 2018-2024 Advanced Micro Devices, Inc. All rights reserved.
```
jkottiku, line 3: Correct year for all the new conf. files. Just 2024 will do.

jkottiku (approved):

PR #904, opened 2025-03-05
Configurable Workgroup count

jkottiku:
Update MI308X-HF iet_stress.conf

### DIFF
--- a/iet.so/src/action.cpp
+++ b/iet.so/src/action.cpp
@@ -98,6 +98,7 @@ using std::fstream;
 
 #define RVS_CONF_BLAS_SOURCE_KEY        "blas_source"
 #define RVS_CONF_COMPUTE_TYPE_KEY       "compute_type"
+#define RVS_CONF_WG_COUNT               "wg_count"
 
 #define IET_DEFAULT_BLAS_SOURCE         "rocblas"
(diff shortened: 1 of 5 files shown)


iet.so/src/action.cpp
```diff
@@ -127,12 +128,13 @@ using std::fstream;
...
 #define IET_DEFAULT_HOT_CALLS           1
 #define IET_DEFAULT_MATRIX_INIT         "default"
-#define IET_DEFAULT_GEMM_MODE        ""
+#define IET_DEFAULT_GEMM_MODE           ""
 #define IET_DEFAULT_BATCH_SIZE          0
 #define IET_DEFAULT_STRIDE_A            0
 #define IET_DEFAULT_STRIDE_B            0
 #define IET_DEFAULT_STRIDE_C            0
 #define IET_DEFAULT_STRIDE_D            0
+#define IET_DEFAULT_WG_COUNT            80
```
manoj-freyr, line 137: 80 is for a specific ASIC ? if that is the case we must put correct value here if no config param is present. hipDeviceProp_t  has maxThreadsPerBlock which is the maximum WG size. Please use this

manoj-freyr (changes requested):

manoj-freyr (approved):

PR #912, opened 2025-03-13
update init method

manoj-freyr:


rvs/conf/MI210/gst_single.conf
```diff
@@ -59,6 +59,7 @@ actions:
   matrix_size_a: 28000
   matrix_size_b: 28000
   matrix_size_c: 28000
+  matrix_init: hiprand
```
jkottiku, line 62: Did you mean **matrix_init: rand** ? Currently, we don't have support for fp32 in hiprand.

jkottiku (approved):

PR #915, opened 2025-03-18
float (fp32) hiprand init.

jkottiku:


src/rvs_blas.cpp
```diff
@@ -1148,6 +1148,32 @@ void rvs_blas::generate_random_matrix_data(void) {
...
+      }
+      else if((ops_type == "sgemm") || (data_type == "fp32_r")) {
+
+        if(hiprandGenerateUniform(hiprand_generator, (float *)da, size_a) != HIPRAND_STATUS_SUCCESS) {
+          std::cout << "\n hiprandGenerateUniform() failed !!!" << "\n";
+          is_error = true;
+          return;
+        }
+
+        if(hiprandGenerateUniform(hiprand_generator, (float *)db, size_b) != HIPRAND_STATUS_SUCCESS) {
```
manoj-freyr, line 1165: Maybe add for which matrix the failure in API happened, so that error logs point at right data.

manoj-freyr (approved):

PR #938, opened 2025-06-05
Add rocBLAS version check for FP8 support

jkottiku:


gst.so/src/action.cpp
```diff
@@ -566,6 +570,16 @@ bool gst_action::get_all_gst_config_keys(void) {
     bsts = false;
   }
 
+#if !defined(RVS_ROCBLAS_HAS_F8_DATATYPES)
```
manoj-freyr, line 573: we must handle in iet also

manoj-freyr (approved):
looks good

PR #954, opened 2025-06-26
MI308X/MI308X-HF fp8 gemm using hipblasLT

jkottiku:
MI308X/MI308X-HF fp8 gemm using hipblasLT
Remove redundant fp8 gemm from gfx950


manoj-freyr (approved):
looks good.

PR #1050, opened 2025-10-28
Enable Lintian Support ROCmValidationSuite

jonatluu:
## Motivation

Install changelog, and copyright to fix lintian errors:
no-copyright-file
debian-changelog-file-missing

## Technical Details

Added function configure_pkg to install changelog, copyright, and lintian override for debian

## Test Plan

Verify lintian errors are fixed and changelog, copyright and lintian overrides are installed

## Test Result

Lintian errors fixed changelog, copyright and lintian overrides are installed

## Submission Checklist

### DIFF
--- a/cmake_modules/utils.cmake
+++ b/cmake_modules/utils.cmake
@@ -145,41 +145,32 @@ function( configure_pkg PACKAGE_NAME_T COMPONENT_NAME_T PACKAGE_VERSION_T MAINTA
 
       # Configure the changelog file
       configure_file(
-        "${CMAKE_SOURCE_DIR}/DEBIAN/changelog.in"
-        "${CMAKE_BINARY_DIR}/DEBIAN/changelog.Debian"
+        "${CMAKE_SOURCE_DIR}/CHANGELOG.md"
+        "${CMAKE_BINARY_DIR}/DEBIAN/CHANGELOG.md"
         @ONLY
       )
 
       if( BUILD_ENABLE_LINTIAN_OVERRIDES )
-	if(DEFINED BUILD_SHARED_LIBS AND NOT ${BUILD_SHARED_LIBS} STREQUAL "")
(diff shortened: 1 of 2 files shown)


cmake_modules/utils.cmake
```diff
@@ -118,3 +118,129 @@ function ( get_version DEFAULT_VERSION_STRING )
     set( VERSION_BUILD  "${VERSION_BUILD}" PARENT_SCOPE )
 
 endfunction()
+
+## Configure Copyright File for Debian Package
+function( configure_pkg PACKAGE_NAME_T COMPONENT_NAME_T PACKAGE_VERSION_T MAINTAINER_NM_T MAINTAINER_EMAIL_T)
+    # Check If Debian Platform
+    find_file (DEBIAN debian_version debconf.conf PATHS /etc)
```
manoj-freyr: check the syntax of find_file.. since you are specifying two files to be checked, they must have a FILES before them right? find_file (DEBIAN NAMES debian_version debconf.conf ..) ?

cmake_modules/utils.cmake
```diff
@@ -118,3 +118,129 @@ function ( get_version DEFAULT_VERSION_STRING )
...
+    if(DEBIAN)
+      set( BUILD_DEBIAN_PKGING_FLAG ON CACHE BOOL "Internal Status Flag to indicate Debian Packaging Build" FORCE )
```
manoj-freyr: we dont have else condition to set it to OFF. what happens in case of non-debian setup ?

jonatluu: This change is just for debian packages

cmake_modules/utils.cmake
jonatluu: I believe the syntax is fine. it follows the pattern 
**find_file (<VAR> name1 [path1 path2 ...])**
you can have multiple names where it will check the path specified 
 
https://cmake.org/cmake/help/latest/command/find_file.html

manoj-freyr (approved):
looks good

PR #1101, opened 2026-02-10
Better code to fetch branch & commit

frepaul:
## Motivation

Branch and commit hash is not fetched properly while running in github runner.
 
## Technical Details
Outside the script commit id is fetched properly, hence use the same as a better option.

## Test Plan
Test it out on the per commit trigger.

## Test Result

NA


frepaul:
Better code already in place.

PR #1106, opened 2026-03-16
github workflow update to upload pacakges to S3

frepaul:
Each time build is done packages created should be uploaded to AWS S3 bucket for PR builds.

This will be updated to upload packages to respective folder based on the branch, nightly & release builds?


frepaul:
Better code is already implemented and merged.

PR #1122, opened 2026-04-09
Add action select (-s) cli. option.

jkottiku:
## Submission Checklist

### DIFF
--- a/rvs/src/rvsexec_do_yaml.cpp
+++ b/rvs/src/rvsexec_do_yaml.cpp
@@ -29,6 +29,9 @@
 #include <iomanip>
 #include <thread>
 #include <fstream>
+#include <vector>
+#include <set>
+#include <sstream>
 
 #include "include/rvsexec.h"
 #include "yaml-cpp/yaml.h"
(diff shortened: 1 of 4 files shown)


rvs/src/rvsexec_do_yaml.cpp
```diff
@@ -81,13 +84,53 @@ int rvs::exec::do_yaml(const std::string& config_file) {
...
+        int idx = std::stoi(token, &pos);
+        if (pos != token.size() || idx < 0)
+          all_numeric = false;
+      } catch (...) {
+        all_numeric = false;
+      }
+    }
+    for (const auto& t : tokens) {
+      if (all_numeric)
+        selected_action_indices.insert(std::stoi(t));
```
manoj-freyr, line 109: this is under assumption that action name is a not a number . Acceptable assumption but cannot validate/guarantee this.

manoj-freyr (approved):

rvs/src/rvsexec_do_yaml.cpp
jkottiku: We will take in either action name or action number as input. If all_numeric is true, its number based else its name based.

yugang-amd (approved):

PR #1126, opened 2026-04-21
pull latest ROCm for schduled job.

frepaul:
This will ensure new ROCm is not causing issue in building a known good RVS mainline.


manoj-freyr (approved):
looks good. scheduled runs pick latest rocm

PR #1127, opened 2026-04-22
update S3 upload to release/nightly

frepaul:
Fixed an issue on upload to release branch even when build is not from release branch. Also simplified the condition check for each type of job and removed artifact upload to github storage.

### DIFF
--- a/.github/workflows/build-relocatable-packages.yml
+++ b/.github/workflows/build-relocatable-packages.yml
@@ -371,7 +371,8 @@ jobs:
     name: Create Release Summary
     needs: [build-ubuntu, build-centos]
     runs-on: ${{ vars.RUNNER_LABEL_UTILITY || 'ubuntu-latest' }}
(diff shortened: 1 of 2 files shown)


.github/workflows/build-relocatable-packages.yml
```diff
@@ -398,84 +374,41 @@ jobs:
     if: github.event_name == 'schedule' || (github.event_name == 'push' && (github.ref == 'refs/heads/master' || github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/release/'))) || (github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository)
```
manoj-freyr: though not done by this PR< do you think we need to add workflow_dispatch too here?Or else, manual runs will not run create-release job right?

manoj-freyr (approved):

.github/workflows/build-relocatable-packages.yml
frepaul: updated to generate always unless cancelled

PR #1129, opened 2026-04-22
update workflow for pkg name/version change

frepaul:
With the new proposal packages generated by RVS need to stick to a new standard to match with compatible ROCm Core version. Install location also changes to /opt/rocm/extras-<ROCm Major>. Inside the install location file layout remain same as before. Within a major release only in-place upgrade with compatibility contract to older and newer ROCm minor releases. Side-by-side install across major version releases.


jkottiku (approved):

manoj-freyr (approved):
lgtm,.

PR #1132, opened 2026-04-24
containerise ubuntu builds

manoj-freyr:
Containerise Ubuntu build job (ubuntu:22.04 docker)

Run the Ubuntu workflow inside an `ubuntu:22.04` container so the build is
reproducible regardless of host

* `ensure_recent_cmake_ubuntu()`: For dependent cmake files calling block(),
  added in CMake 3.25 but Ubuntu 22.04 ships 3.22. Try Kitware APT repo first,
  fall back to pip; gated by `RVS_SKIP_KITWARE_REPO` for air-gapped runners.
* Mark `$GITHUB_WORKSPACE` as `safe.directory` - container UID differs from
  the checkout owner so git rev-parse otherwise yields "0000000".
* In the Ubuntu/Debian branch, point `CMAKE_CXX_COMPILER` at hipcc and set
  `GCC_TOOLCHAIN=/usr` when /usr/include/c++/*/barrier exists, mirroring the
  `--gcc-toolchain` plumbing already used for RHEL/AlmaLinux. Without this
  hipcc (clang) cannot find C++20 <barrier> against the system libstdc++.

### DIFF
--- a/.github/workflows/build-relocatable-packages.yml
+++ b/.github/workflows/build-relocatable-packages.yml
@@ -48,18 +48,36 @@ jobs:
   build-ubuntu:
     name: Build Ubuntu Packages
     runs-on: ${{ vars.RUNNER_LABEL || 'ubuntu-22.04' }}
+    container:
+      image: ubuntu:${{ matrix.ubuntu_version }}
     strategy:
       matrix:
         ubuntu_version: ['22.04']
+    # Apply to every step in the job so apt never drops into an interactive
+    # frontend (tzdata, libssl, etc.) — each Actions step runs in a fresh
+    # shell, so an `export` inside one step does not propagate to the next.
+    env:
+      DEBIAN_FRONTEND: noninteractive
+      TZ: Etc/UTC
     outputs:
       s3_bucket_ubuntu: ${{ steps.upload-s3-ubuntu.outputs.bucket }}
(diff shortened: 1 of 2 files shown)


.github/workflows/build-relocatable-packages.yml
```diff
@@ -47,19 +47,37 @@ env:
 jobs:
   build-ubuntu:
     name: Build Ubuntu Packages
-    runs-on: ${{ vars.RUNNER_LABEL || 'ubuntu-22.04' }}
+    runs-on: ${{ vars.RUNNER_LABEL_CONTAINER || 'ubuntu-latest' }}
```
frepaul: why latest, cant we keep it to ubuntu:22.04?

manoj-freyr: it is just to pin the fall back VM on GH runner. we still have 22.04 docker container fixed for our purposes if you see the ubuntu:${{ matrix.ubuntu_version }} part we pin it to 22.04. What I ask for latest is the docker host(in case of GH actions it would be a VM ) and latest is the default of GH docker hosts and nothing more.. vars.RUNNER_LABEL_CONTAINER is still the main mover and the default is just a safe public-runner fallback.

manoj-freyr: defaulted to 22 and also to right variable

frepaul (approved):

PR #1133, opened 2026-04-24
rvs version print should match package version

frepaul:
rvs --version should print the same version MAJOR.MINOR.PATCH atleast matching to package.


manoj-freyr (approved):
lgtm

PR #1134, opened 2026-04-24
fetch path and version relative to install

frepaul:
## Motivation

RVS prints ROCm varsion it links/loads.Undate the version fetch using API provided by ROCm.If that option is turned OFF fallback to ROCM_PATH environment variable.

For all the module/config reading inside RVS, use self discovery of RVS location(linux only) and use path relative to executable This is to ensure relocation of RVS possible.

For both ROCm path & RVS path last choice will be to use compile time backed in path.


jkottiku:
Hope these changes are backward compatible with old ROCm folder structure ? Package and source code build and execution method ?

jkottiku (approved):

PR #1135, opened 2026-04-27
Update install instruction, RPATH for RVS.

frepaul:
Since TGZ based install cannot manage dependencies and setup env, user need to know what are the pre-install and post install checklist that need to be done to make rvs functional.

RVS need libomp from llvm/lib provided in ROCm to work. Hence add RPATH/RUNPATH to ROCm_path/lib and ROCm_path/lib/llvm/lib. Anytime ROCm is installed in a non-standard location using TGZ user shall be instructed to set LD_LIBRARY_PATH with appropriate path

Also cleaned up build RPATH if built from github and keep it only in the case of local build(using CMAKE_SKIP_BUILD_RPATH).

### DIFF
--- a/.github/workflows/README_BUILD_PACKAGES.md
+++ b/.github/workflows/README_BUILD_PACKAGES.md
@@ -346,16 +346,69 @@ sudo rpm -i --replacefiles --nodeps amdrocm7-rvs-*.rpm
 
 ### Any Linux Distribution (TGZ - Relocatable)
 
+#### Pre-install: ROCm
+
+Before you install the RVS TGZ, **ROCm must be installed, configured, and on your `PATH` / `LD_LIBRARY_PATH` as in AMD’s documentation** so HIP, HSA, and other ROCm libraries are discoverable. Follow the current Linux install guide in **rocm docs**:
+
+- **ROCm documentation (start here)**: <https://rocm.docs.amd.com/>
+- **Linux install / deployment (paths, env, post-install)**: <https://rocm.docs.amd.com/en/latest/deploy/linux/index.html>
+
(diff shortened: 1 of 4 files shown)


atnair-amd (approved):

.github/workflows/README_BUILD_PACKAGES.md
```diff
@@ -346,16 +346,69 @@ sudo rpm -i --replacefiles --nodeps amdrocm7-rvs-*.rpm
 
 ### Any Linux Distribution (TGZ - Relocatable)
 
+#### Pre-install: ROCm
+
+Before you install the RVS TGZ, **ROCm must be installed, configured, and on your `PATH` / `LD_LIBRARY_PATH` as in AMD’s documentation** so HIP, HSA, and other ROCm libraries are discoverable. Follow the current Linux install guide in **rocm docs**:
+
+- **ROCm documentation (start here)**: <https://rocm.docs.amd.com/>
+- **Linux install / deployment (paths, env, post-install)**: <https://rocm.docs.amd.com/en/latest/deploy/linux/index.html>
```
peterjunpark: This link will need to be updated to the ROCm installation instructions with TheRock. 

The 7.13 instructions won't be available until the release is generally available, but the url will probably look like https://rocm.docs.amd.com/en/latest/install/rocm.html

The page for 7.12 (preview) is here
https://rocm.docs.amd.com/en/7.12.0-preview/install/rocm.html

frepaul: updated with "https://rocm.docs.amd.com/en/latest/install/rocm.html"

frepaul: Actually I prefer to keep the working link and later update if the link changes..

manoj-freyr (approved):

PR #1137, opened 2026-04-28
Iterations (hot calls) based GST execution.

jkottiku:
## Submission Checklist

### DIFF
--- a/gst.so/src/gst_worker.cpp
+++ b/gst.so/src/gst_worker.cpp
@@ -202,10 +202,10 @@ bool GSTWorker::do_gst_ramp(int *error, string *err_description) {
 
   // make sure that the ramp_interval & duration are not less than
   // NMAX_MS_GPU_RUN_PEAK_PERFORMANCE (e.g.: 1000)
-  if (run_duration_ms < NMAX_MS_GPU_RUN_PEAK_PERFORMANCE)
+  if (run_duration_ms > 0 && run_duration_ms < NMAX_MS_GPU_RUN_PEAK_PERFORMANCE)
     run_duration_ms += NMAX_MS_GPU_RUN_PEAK_PERFORMANCE;
 
-  if (ramp_interval < NMAX_MS_GPU_RUN_PEAK_PERFORMANCE)
+  if (ramp_interval > 0 && ramp_interval < NMAX_MS_GPU_RUN_PEAK_PERFORMANCE)
     ramp_interval += NMAX_MS_GPU_RUN_PEAK_PERFORMANCE;
 
   // stage 1. setup rvs blas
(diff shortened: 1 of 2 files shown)


gst.so/src/gst_worker.cpp
```diff
@@ -621,71 +621,34 @@ void GSTWorker::run() {
...
+    action.action_callback(&action_result);
+
+    return;
   }
 
-  log_interval_gflops(max_gflops);
   check_target_stress(max_gflops);
 }
 
 void GSTWorker::log_gst_test_result(bool gst_test_passed) {
```
manoj-freyr: if unused we can fully remove this instead of thi no-op right?

gst.so/src/gst_worker.cpp
```diff
@@ -621,71 +621,34 @@ void GSTWorker::run() {
...
-    }
+    action_result.state = rvs::actionstate::ACTION_COMPLETED;
+    action_result.status = rvs::actionstatus::ACTION_FAILED;
+    action_result.output = msg.c_str();
+    action.action_callback(&action_result);
+
+    return;
   }
 
-  log_interval_gflops(max_gflops);
```
manoj-freyr, line 647: is this removal intentional ?

gst.so/src/gst_worker.cpp
```diff
@@ -498,7 +498,7 @@ bool GSTWorker::do_gst_stress_test(int *error, std::string *err_description) {
     log_interval_milliseconds = time_diff(gst_end_time,
         gst_log_interval_time);
 
-    if (log_interval_milliseconds >= log_interval && num_gemm_ops > 0) {
+    if ((log_interval_milliseconds >= log_interval || 0 == run_duration_ms) && num_gemm_ops > 0) {
```
manoj-freyr: one nitpick but possible scenario. We have duration and log interval as 0 and have iters as 1. this condtional will pass and once we enter the conditional, and time_diff() above , for mi3xx and above can definitely give a 0 return value if duration taken is less than 1ms. this will make seconds as 0 and thus wont enter the below line at 505 and we will have max_gflops as 0(, since it wont be updated. and when we can check_target_stress() it will fail the test. 
Please use same iter check or max glops being non zero when you call check_target_stress() later

manoj-freyr (changes requested):

gst.so/src/gst_worker.cpp
jkottiku: Yes, this is an duplicate logging thats redundant.

gst.so/src/gst_worker.cpp
jkottiku: In the scenario, log_interval is 0 and not log_interval_milliseconds right ? So, seconds_elapsed will have non-zero value and test will pass. Also tested and verified.

manoj-freyr: seconds_elapsed  will be non zero if log_interval_milliseconds also return zero. time_diff() returns values in millisecond resolutions right? if so, anything less than 1ms, saay even 900microseconds, will return 0 and 0 is >= 0, this is possible right in mi3xx or better systems?

jkottiku: In that case, better move the resolution in microseconds instead of milliseconds which should handle this edge condition.

manoj-freyr (approved):
looks good, comments addressed.

PR #1157, opened 2026-05-20
Fixed LD Library Path

urtiwari:
RVS failed to run as the LD library path was not right


frepaul (approved):
Ok with this change, but I would recommend splitting the tests orchestration and workflow into two sperate files.

PR #1173, opened 2026-06-01
fix typerror on local runner with older python

frepaul:
## Motivation

Fix typeerror with local runner running on python-3.8

## Technical Details

TypeError: 'type' object is not subscriptable on line 6 of the inline script is from list[str] annotations. That syntax needs Python 3.9+. Removing Python 3.9+ type hints (list[str]) the self-hosted runner uses Python 3.8.

## Test Plan

Functional active branch nightly running on local runner.
## Test Result

NA


manoj-freyr (approved):
looks good

PR #1175, opened 2026-06-01
use std filesystem call instead of mkdir

frepaul:
## Motivation

Executing commands have potential security loop hole.

## Technical Details

using std filesystem call is more safer than using system commands to create folder for rvs log

## Test Plan

RUN rvs and generate log

## Test Result

NA


frepaul:
creating a new one.

PR #1178, opened 2026-06-02
Fix broken links in header & footer

yugang-amd:
## Submission Checklist

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -23,7 +23,7 @@ exclude_patterns = [
 # for PDF output on Read the Docs
 project = "RVS Documentation"
 author = "Advanced Micro Devices, Inc."
-copyright = "Copyright (c) 2023-2025 Advanced Micro Devices, Inc. All rights reserved."
+copyright = "Copyright (c) 2023-2026 Advanced Micro Devices, Inc. All rights reserved."
 version = version_number
 release = version_number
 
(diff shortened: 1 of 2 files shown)


docs/conf.py
```diff
@@ -30,6 +30,21 @@
 html_title = left_nav_title
 extensions = ["rocm_docs"]
 html_theme = "rocm_docs_theme"
-html_theme_options = {"flavor": "rocm"}
+html_theme_options = {
+    "announcement": f"This is ROCm 7.13 technology preview release documentation. For the latest production stream release, refer to <a id='rocm-banner' href='https://rocm.docs.amd.com/en/latest/'>ROCm documentation</a>.",
```
peterjunpark: We can probably remove the announcement banner. RVS is an "extra" and can have a different release cadence from ROCm Core SDK releases.

docs/conf.py
```diff
@@ -30,6 +30,21 @@
...
+    "flavor": "generic",
+    "header_title": f"ROCm Extras",
+    "header_link": f"https://rocm.docs.amd.com/en/7.13.0-preview/components/extras.html",
+    "version_list_link": f"https://rocm.docs.amd.com/en/7.13.0-preview/release/versions.html",
```
peterjunpark: I think we should either remove the version list or create a decoupled one for RVS

yugang-amd: Removed the list for now...

alexxu-amd:
7.2.3 and all past versions are still using rocm standard flavor. Have you checked with Amit/Ram if it is OK to be inconsistent in this case to resolve the deadlink?

frepaul (approved):

yugang-amd:
> 7.2.3 and all past versions are still using rocm standard flavor. Have you checked with Amit/Ram if it is OK to be inconsistent in this case to resolve the deadlink?

Ram/Amit would like RVS to follow similar patterns as we have for Optiq: [What is ROCm Optiq? — ROCm Optiq (Beta) 0.4.0 documentation](https://rocm.docs.amd.com/projects/roc-optiq/en/latest/what-is-optiq.html)

alexxu-amd (approved):

PR #1199, opened 2026-06-08
Build and install TransferBench CLI as part of RVS

saadrahim:
## Motivation

[ROCm/TransferBench](https://github.com/ROCm/TransferBench) is already an indirect dependency of RVS — PEBB and PBQT consume its headers via a configure-time clone in `CMakeTransferBenchDownload.cmake`. The CLI binary, however, is not built or shipped, so users have to install TransferBench separately to get the same standalone tool the rvs modules are built against.

This PR vendors TransferBench as a git submodule (replacing the configure-time clone) and builds & installs its CLI binary as part of the rvs package, so users get a single, version-matched install.

## Technical Details

- Add `external/TransferBench` as a git submodule, pinned to `c78c4ae` (TransferBench v1.66.02 — the same SHA `CMakeTransferBenchDownload.cmake` was previously cloning, so PEBB/PBQT see byte-identical headers).
- Delete `CMakeTransferBenchDownload.cmake` and its `configure_file`/`execute_process` block in `CMakeLists.txt`. `TRANSFERBENCH_INC_DIR` now points at `external/TransferBench/src/header`. A fatal-error check tells the user to run `git submodule update --init --recursive` if the submodule isn't initialised.
- Add `CMakeTransferBenchCLI.cmake`: an `ExternalProject_Add` sub-build that invokes TransferBench's own CMake. TransferBench pins `amdclang++` as `CMAKE_CXX_COMPILER` *before* its `project()` call, so `add_subdirectory` would fight the parent compiler — `ExternalProject_Add` isolates the two builds.
- Install the resulting `TransferBench` binary with `COMPONENT rvsmodule` (the same component as `rvs`), so it lands in the same DEB and RPM as the rest of RVS at `${CPACK_PACKAGING_INSTALL_PREFIX}/${CMAKE_INSTALL_BINDIR}/TransferBench`.
- Add `BUILD_TRANSFERBENCH_CLI` cmake option (default `ON`) for opt-out on toolchains that can't satisfy TransferBench's `cmake_minimum_required(VERSION 3.16)`.
- Add `submodules: recursive` to all four `actions/checkout@v4` steps in `.github/workflows/rvs-nightly-tests.yml`. `build-relocatable-packages.yml` already passed it.
- README updated to use `git clone --recurse-submodules` plus a fallback for already-cloned repos. CHANGELOG updated under RVS 1.5.0 → Added.

## Test Plan

```bash
git clone --recurse-submodules https://github.com/ROCm/ROCmValidationSuite.git
cd ROCmValidationSuite
cmake -B build -DROCM_PATH=/opt/rocm -DCMAKE_INSTALL_PREFIX=/opt/rocm \
  -DCPACK_PACKAGING_INSTALL_PREFIX=/opt/rocm
cmake --build build -j
sudo cmake --install build
```

Then verify:

1. `TransferBench --help` runs from the install bindir.
2. `rvs -m pebb` and `rvs -m pbqt` still configure and run (headers come from the submodule at the same pinned SHA).
3. `cd build && cpack -G DEB && cpack -G RPM` — inspect the resulting packages with `dpkg -c` / `rpm -qlp` and confirm `bin/TransferBench` is included.
4. Opt-out path: re-configure with `-DBUILD_TRANSFERBENCH_CLI=OFF`, rebuild, confirm the CLI is absent but PEBB/PBQT still build.

## Test Result

Local build on <fill in env>: rvs + TransferBench CLI both built and installed; PEBB/PBQT regressions clean; DEB and RPM both contain `bin/TransferBench` alongside `bin/rvs`.

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.


frepaul (approved):
Looks good. Need to pickup TransferBench release tag. Same can we done to pickup the latest release tag as part of github action workflow code checkout.

peterjunpark (approved):

manoj-freyr (commented):
--recurse-submodules , this necessity might affect exisiting users who just do git clone. and also defaukt can be OFF and can be turned on as needed.

PR #1235, opened 2026-06-24
Edited the nightly run to only run when schedule and on PUSH

urtiwari:
Edited the nightly run to only run when schedule and on PUSH

### DIFF
--- a/.github/workflows/rvs-nightly-tests.yml
+++ b/.github/workflows/rvs-nightly-tests.yml
@@ -16,17 +16,18 @@ name: RVS Nightly Tests
 #
 # Triggers
 # --------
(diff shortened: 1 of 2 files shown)


.github/workflows/rvs-nightly-tests.yml
```diff
@@ -16,17 +16,18 @@ name: RVS Nightly Tests
...
     - cron: '0 15 * * *'
-  workflow_run:
-    workflows: ["Build Relocatable Packages"]
-    types:
-      - completed
+  push:
+    branches:
+      - master
+      - main
+      - 'release/**'
```
frepaul: Why do you need this for release/**

urtiwari: Removed release, It was part of push so got added for release as well

frepaul (approved):

PR #1241, opened 2026-06-26
fix fragile steps in build script.

frepaul:
For some reason finding GCN path from ROCm is causing issue on github runner. Trying a different approach and do error reporting for future debug.


urtiwari (approved):
LGTM.

PR #1242, opened 2026-06-28
ignore downloading -test- ROCm package

frepaul:
ROCm 7.15 onwards added tarball with tests as package with -tests- added to the name before version. Current logic downloads that and fail to fetch the ROCm version from the downloaded tarball. With this change package that has name "-tests-" will be ignored.

No ROCm download progress print in github action and print success.


urtiwari (approved):
LGTM.

PR #1244, opened 2026-07-06
AMD SMI was initialized multiple times

manoj-freyr:
## Motivation

Fix — centralized AMD SMI lifecycle. One ownership and idempotent cleanup API to work with both CLI run as well as API and ensure multi module loads are handled (ufor -r and -m cases)

## Technical Details

amdsmi_init() called more than once in a process as of now and a single shutdoqwn() call at exit of a RVS module. This resulted in corrupted heap and causes a segfault. A mutex guard for amd smi as a resource ensures single init and cleanup path . This solves cases for : 

-  Multi module call in a single process: like -m or -r runs and also a conf file with many modules in it.
- Single action runs
- API based calls. 

## Rootcause
Typical RVS run is : 
```

- First init — gst module rvs_module_init() → gpulist::Initialize()
- Test runs and passes; action_destroy() cleans up rvs_blas
- Second init — end-of-run summary → systemOverview() → gpulist::Initialize() again 
- Single shutdown — module::terminate() → gst rvs_module_terminate() → amdsmi_shut_down()
```
This double init and single shutdown causes Clib heap corruption error. 
WIth the current PR: we use guard to init and shutdown smi. We ensure gpulist::Initialize() calls smi init and gpulist::Terminate() calls smi shutdown. gpulist::Terminate() is called from module::terminate() after unloading modules. 
For API cases, we add module::terminate() in rvs_terminate()   but since this is idempotent and a no-op in case SMI already shutdown.
Retaubed existing behavior in unit tests as they run as standalone binaries.
## Test Plan

Run rvs with a multi module conf on latest rocm,.

### DIFF
--- a/rvs/src/rvs_interface.cpp
+++ b/rvs/src/rvs_interface.cpp
@@ -27,6 +27,7 @@
 #include <include/rvs.h>
 #include <include/rvsinternal.h>
 #include <include/rvsexec.h>
+#include <include/rvsmodule.h>
 #include <map>
 #include <mutex>
 
@@ -291,6 +292,10 @@ rvs_status_t rvs_terminate(void) {
     return RVS_STATUS_INVALID_STATE;
   }
   rvs_state = RVS_STATE_UNINITIALIZED;
+
+  /* Ensure AMD SMI and loaded modules are torn down if the client skipped
+   * session destroy or exited without running module::terminate via exec. */
+  rvs::module::terminate();
 
   return RVS_STATUS_SUCCESS;
 }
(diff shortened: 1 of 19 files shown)


rvs/src/rvs_interface.cpp
```diff
@@ -292,6 +293,10 @@ rvs_status_t rvs_terminate(void) {
   }
   rvs_state = RVS_STATE_UNINITIALIZED;
 
+  /* Ensure AMD SMI and loaded modules are torn down if the client skipped
+   * session destroy or exited without running module::terminate via exec. */
+  rvs::module::terminate();
```
jkottiku: Is this really required here ? As it would be called in run() function anyways.

jkottiku:
Still we call rvs::gpulist::Initialize() from modules right ? Should we move it to rvs::exec::run() function as well ?

PR #1270, opened 2026-07-29
Adding the files to run RVS pipeline in docker

urtiwari:
Added the capability to run RVS nightly in docker


urtiwari:
The following are the changes done,

Pick a RVS tar (*-Linux.tar.gz with -r0715.20260724 style name).
Build a Docker image with the same ROCm version as that tar (from multiarch).
Download the RVS tar, extract it in the container.
Run rvs -r 4 inside Docker with GPU access.

frepaul (approved):

PR #1276, opened 2026-07-31
Fix PBQT peer-only mode JSON GPU IDs and action result.

manoj-freyr:
1. Use gpu_id[i] instead of gpu_id[srcnode] when logging peer connectivity JSON so srcgpu matches text logs on systems where HSA node IDs differ from enumeration indices.
2. Add MI350P pbqt/pebb configs,

### DIFF
--- a/rvs/conf/MI350P-450W/pbqt_single.conf
+++ b/rvs/conf/MI350P-450W/pbqt_single.conf
@@ -1,6 +1,6 @@
 # ################################################################################
 # #
-# # Copyright (c) 2025-2026 Advanced Micro Devices, Inc. All rights reserved.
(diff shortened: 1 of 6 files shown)


rvs/conf/MI350P-450W/pbqt_single.conf
```diff
@@ -0,0 +1,118 @@
+# ################################################################################
+# #
+# # Copyright (c) 2025-2026 Advanced Micro Devices, Inc. All rights reserved.
```
jkottiku: Just "2026" as it a new file.

jkottiku (changes requested):

jkottiku (approved):

PR #1294, opened 2026-08-19
Set Default GPU_FAMILY to multiarch

urtiwari:
Default GPU_FAMILY is multiarch (was gfx110X-all) in the Dockerfile and build-rocm-image.sh.
The docker nightly workflow sets GPU_FAMILY to vars.RVS_DOCKER_GPU_FAMILY or multiarch.


urtiwari:
There was an issue with this PR so closing.

PR #1302, opened 2026-08-26
update permission/path check for local build

frepaul:
allow group-writable .so files when you own them (local builds). Still reject world-writable files; when running as root, still require root-owned, non-group-writable modules.

search build/bin/ first (dev layout),
then ../lib/rvs/ (install layout), then other fallbacks.

Better error reporting — prefer a specific verification error over a generic "could not resolve module path" when both occur.

### DIFF
--- a/include/rvs_util.h
+++ b/include/rvs_util.h
@@ -86,9 +86,9 @@ std::string rvs_get_rvs_modules_lib_dir_string(void);
 /**
  * @brief Validate a module .so before dlopen().
  *
- * On Linux: regular file, not group/world writable. Ownership is checked only
- * for non-root callers (must be owned by the effective user or root). When
- * euid is 0 (root or sudo), any owner is allowed if permissions pass.
+ * On Linux: regular file, not world-writable; group-writable only when owned by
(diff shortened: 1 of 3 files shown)


urtiwari (commented):
Stale comment in src/rvs_util.cpp (should fix)
The block comment above rvs_verify_module_so_for_dlopen() still says the file "is not group- or world-writable". That no longer matches the implementation or the updated header in rvs_util.h. Update the .cpp comment so docs stay in sync.

2. Stale comment in rvsmodule.cpp (nit)
rvs_dlopen_verified_module() still references "all three module search paths"; there are now four. Minor, but worth correcting.

3. Non-root group-writable check is redundant but confusing
if ((st.st_mode & S_IWGRP) && st.st_uid != euid) {
  *err_msg = "module is group-writable";
  return false;
}
After the preceding owner check, this branch only runs when st_uid == 0 (root-owned, group-writable, non-root caller). The message "module is group-writable" is accurate but doesn't convey why a root-owned file is rejected here. A one-line comment explaining that case, or a slightly more specific error string, would help future readers.

manoj-freyr (approved):

PR #1312, opened 2026-09-01
config addition for RX 9050

manoj-freyr:
enable configs for IET and GST for RX 9050


jkottiku:
Instead of iet_single.conf, shall we just have iet_stress.conf with single action to test max. power?

jkottiku (approved):

PR #1324, opened 2026-09-15
prepare nightly for signing CI.

frepaul:
## Motivation

Nightly RVS packages are built unsigned. A separate signing CI needs a predictable input prefix from the scheduled build. Copies of built artifacts under nightly/unsigned/
(scheduled default branch).


urtiwari (commented):
Blocking
1. reprepro includedeb will fail when the nightly version does not change

for deb in $DEBS; do
 echo "Including $(basename "$deb") in suite stable ..."
 reprepro -b "$STAGING" includedeb stable "$deb"
If amdrocm*-rvs is already in stable at the same version (typical when CPack version is the git tag, not a date), reprepro exits non-zero and the Ubuntu job fails. Handle replace explicitly, for example remove the old package first, or use reprepro’s overwrite/same-version ignore flags, and decide whether old versions stay in pool/ for signing CI.

2. Unrelated CMake change in a packaging PR

The CMAKE_MODULE_PATH prepend is a real TheRock/find_package fix, but it is not “prepare nightly for signing CI.” It changes every local and CI configure. Split it so a CMake regression cannot block, or hide behind, the S3 work.

High
3. Unsigned prefixes accumulate; signing CI has no “latest” pointer

aws s3 sync has no --delete. RPM/TGZ copies are additive. DEB pool/ grows. The README says signing jobs discover packages from the tree or an S3 listing and that this workflow does not write a run manifest.

Signing CI can then pick an old .deb/.rpm next to today’s. Either:

keep only the current N versions (and --delete stale objects), or
write a small manifest (latest.json / run id) under nightly/unsigned/ as part of this PR, not later.
4. Docs placeholders lost the bucket name

The new apt/yum examples render as https://.s3.amazonaws.com/... and s3:// /nightly/unsigned/. That is unusable as copy-paste. Use a visible placeholder such as <AWS_S3_BUCKET>.

5. First unsigned DEB job can fail on reprepro missing even though you installed it

You added reprepro to the Ubuntu bootstrap. The script treats it as fatal (::error:: + exit 1), which is correct. Confirm that job uses the same bootstrap path on scheduled default-branch runs (container vs runner). If a matrix entry skips bootstrap, nightly Ubuntu goes red.

urtiwari (commented):
Blocking
1. Missing metadata is a successful skip, which desyncs signing input

rvs-deb-unsigned-repo.sh exits 0 if there are no .debs (no deb.json). upload-rpm-tar-unsigned can sync an empty staging with --delete and still succeed. rvs-unsigned-publish-latest.sh then:

if ! aws s3 cp .../deb.json ...; then
  echo "::notice::... skipping latest.json"
  exit 0
fi
Same for rpm-tar.json.

On a scheduled default-branch run that is the contract for signing CI:

rpm/ and tar/ may already have been replaced (--delete).
latest.json stays on yesterday’s keys (or is never written).
Signing CI downloads objects that no longer exist, or signs the previous night.
For schedule + default branch, missing packages or missing fragments should fail the step. exit 0 is only reasonable when unsigned publish is not supposed to run.

2. New OIDC step does not mask AWS secrets

publish-unsigned-latest copies assume-role into GITHUB_ENV but omits the ::add-mask:: lines used in the Ubuntu/CentOS jobs. Session token and secret key can leak in logs. Match the existing masking (and keep id-token: write, which is already workflow-wide).

High
3. Empty RPM/TGZ glob + --delete wipes the unsigned prefix

for f in ./build/amdrocm*-rvs*.rpm; do
  [ -f "$f" ] || continue
  cp "$f" "$RPM_STAGING/"
done
aws s3 sync "$RPM_STAGING/" "s3://.../nightly/unsigned/rpm/" --delete
If the glob matches nothing, staging is empty and S3 rpm/ (and likewise tar/) is deleted. The later cp ... || true + createrepo path can publish an empty repo and still upload rpm-tar.json without rpm/tar keys. Require at least one RPM (and one TGZ if that is part of the handoff) before --delete.

4. reprepro remove stable "$pkg" "$ver" does not do what the README says

reprepro remove <codename> <package-names> — the version is treated as another package name. The archive is also empty (no S3 merge), so remove never runs against last night’s pool. Replacement is entirely --delete on sync. Drop the remove line or the docs will send the next person down the wrong path.

urtiwari (approved):

manoj-freyr (approved):

PR #1326, opened 2026-09-15
always set cmake_module_path

frepaul:
With find_package() of various module potentially adding CMAKE_MODULE_PATH, there is a chance that with the current code path may not be set and cannot find the utils module.


manoj-freyr (approved):
looks good

PR #1330, opened 2026-09-21
accumulate unsigned nightly packages

frepaul:
S3 ARN do not have permission to delete objects. Accumulate files for signing and later stop uploading to nightly/rvs. New ARN may help cleanup S3 later.

Fix issue with JSON upload issue, installing awscli using pip


urtiwari (commented):
Blocking
1. The pip awscli install checks aws before PATH is updated

publish-unsigned-latest is the job that uploads latest.json. On ubuntu-latest, ~/.local/bin is not on PATH in that same step. $GITHUB_PATH applies only to later steps, so aws --version at the end of the install step fails and the job never reaches the upload.

python3 -m pip install --user --break-system-packages awscli \
  || python3 -m pip install --user awscli
echo "$HOME/.local/bin" >> "$GITHUB_PATH"
aws --version
Export it in this step as well:

export PATH="$HOME/.local/bin:$PATH"
aws --version
2. A failed S3 download is treated as an empty archive, then written back

aws s3 sync "s3://${BUCKET}/${DEB_PREFIX}/conf/" "$STAGING/conf/" --no-progress 2>/dev/null || true
aws s3 sync "s3://${BUCKET}/${DEB_PREFIX}/pool/" "$STAGING/pool/" --no-progress 2>/dev/null || true
aws s3 sync "s3://${BUCKET}/${DEB_PREFIX}/dists/" "$STAGING/dists/" --no-progress 2>/dev/null || true
A 403 or a network error looks like "no archive." reprepro then builds a fresh conf/db and dists/ for tonight's .deb only, and aws s3 sync without --delete overwrites those keys. Older pool objects remain, but they are no longer in the reprepro database, so later nights will not pick them up. That is a permanent loss of the accumulated apt archive.

The same || true is on the unsigned RPM download. That one is recoverable the next night, because createrepo rereads whatever RPMs are still in the prefix. The DEB database is not.

Fail the job if sync fails. An empty prefix on the first run is success with no files, not a non-zero exit.

High
3. This run's deb.json can record an old pool object

After includedeb, the script looks up the pool path like this:

pool_rel=$(find "$STAGING/pool" -name "$(basename "$deb")" | head -1 || true)
if [ -z "$pool_rel" ]; then
  pool_rel=$(find "$STAGING/pool" -name "${pkg}_*.deb" | head -1 || true)
fi
find order is undefined. If reprepro's pool filename differs from the CPack filename, ${pkg}_*.deb matches every historical version, and head -1 can write last night's key and digest into deb.json / latest.json. Signing CI would then fetch the wrong object.

Resolve the path from the package and version you just included (reprepro's pool name for that version), and fail if the number of resolved paths is not the number of .deb files. -s "$NEW_META" only checks that at least one line was written.

reprepro removefilter for the same Package+Version, then includedeb, is the right way to overwrite that object with PutObject only. Drop || true on removefilter so a failed remove does not fall through into a confusing includedeb error.

urtiwari (approved):

